PR #1, opened 2016-08-19
Dial down smokescreen logging.

gphat:
# What's this PR do?

Dials down smokescreen logging, since it's pretty loud.

Specifically, it sets the proxy's verbose flag to false. That control's the [rudimentary logging](https://github.com/elazarl/goproxy/blob/bee2110779c23d999b253c0ec8873757dc43d9a3/ctx.go#L56) and prevents a bunch of stuff from coming out. Specifically, `WARN` will continue to work.

# Why?

Specifically, it's been responsible for about 115MM log events in the last 24 hours in our setup.

# Note

I'm biasing to action here. Not sure what uses these logs and if they are needed for anything. Happy to adjust!

r? @antifuchs 
cc @andrew-d @zenazn @ebroder

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -101,7 +101,7 @@ func errorResponse(req *http.Request, err error) *http.Response {
 
 func buildProxy() *goproxy.ProxyHttpServer {
 	proxy := goproxy.NewProxyHttpServer()
-	proxy.Verbose = true
+	proxy.Verbose = false
 	proxy.Tr.Dial = dial
 	proxy.OnRequest().DoFunc(func(req *http.Request, ctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
 		ctx.Logf("Received HTTP proxy request: "+


andrew-d:
This sounds like a good idea to me; I could be convinced that we want to keep some of [the log messages](https://github.com/stripe/smokescreen/search?utf8=%E2%9C%93&q=.Logf), but I don't believe any of them are important enough that we should care right now 😛

zenazn:
These are the only logs we have to help us debug connectivity issues on outbound webhooks. I agree the current logs are quite... chunderous, but I'd hate for the result of this to be that we turn off smokescreen logging of any sort and start flying blind 🙈 

What plans do we have to make smokescreen logs great ~~again~~?

antifuchs:
@zenazn, I don't think that's true - we do track and alert based on error rates for failures using our datadog setup; I get that we'd want it for debugging, but we don't realllly need it for knowing the service is up. Besides, the log messages aren't all that helpful & repetitive a lot of times.

antifuchs:
Given our non-log-based metrics collection, I'm
👍 
on merging.

zenazn:
Sorry—I specifically meant flying blind when debugging

antifuchs:
Ah, I mean, feel free to re-enable verbosity when you have to figure something out? For that purpose, the `Verbose` flag could be a command-line arg (defaulting to off)...

gphat:
I'm happy to adjust the PR to make that a thing! Would that be helpful @zenazn?

zenazn:
I don't particularly enjoy `smokescreen`'s current logging; I was suggesting that we take some time (not necessarily now) to write logging that is both dense and helpful (and e.g., includes `action_id`s, and is easy to Splunk and ???)

Post-hoc verbose flags ~never work—they're never on when they're most helpful, and since they're ~never on, they tend to bitrot very quickly. I would be more in favor of eliminating logs all together and accepting the consequences there than adding a post-hoc verbose flag.

gphat:
@zenazn It's unclear to me what action you'd like here. Are you saying we should *not* merge this in favor of revamping logging or that we can merge this and it would be nice to spend time on it in the future. Or perhaps something else?

It should be noted that the underlying dep is where this trimming is happening now. If smokescreen itself should improve it's logging then that's likely a different PR and almost certainly a different person.

zenazn:
I'm realistically hoping for a merge + an assigned jira ticket (although an unassigned one "fine" too). I'd hate for us to merge this and to forget about writing good logs entirely.

gphat:
Makes sense! I made a ticket to track this. If you'd prefer it be in this (public) repo, lemme know! I'm happy to add.

Thanks everyone! I'll hold off on merging until I can talk to someone about deployment since this one is new to me.

PR #2, opened 2017-01-06
Add logging for every completed request

julia-stripe:
r? @rb-stripe @dougbarth-stripe 

We want to be to log the IPs the proxy is connecting to. This adds a log line that logs that and several other things. what exactly it logs is probably best explained by this example log line:

>  2017/01/06 10:29:02 Completed response: src_host="::1" src_port=54211 host="example.com" dest_ip="93.184.216.34" dest_port=80 start_time=1483727342 end_time=1483727342 content_length=1270

@rb-stripe suggested logging the length of the body, but if we want to have everything on the same log line I'm pretty sure we can only log the value of the Content-Length header. I think the only place the actual value of that is accessible is here (goproxy logs it): https://github.com/elazarl/goproxy/blob/master/proxy.go#L139-L143.

Also maybe we should log requests in addition to / instead of responses.

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -109,6 +109,7 @@ func buildProxy() *goproxy.ProxyHttpServer {
 			ctx.Req.RemoteAddr,
 			ctx.Req.Host,
 			ctx.Req.RequestURI)
+		ctx.UserData = time.Now().Unix()
 		return req, nil
 	})
 	proxy.OnRequest().HandleConnectFunc(func(host string, ctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
@@ -122,9 +123,12 @@ func buildProxy() *goproxy.ProxyHttpServer {
 			ctx.Resp = errorResponse(ctx.Req, err)
 			return goproxy.RejectConnect, ""
 		}
+		ctx.UserData = time.Now().Unix()
 		return goproxy.OkConnect, resolved
 	})
+
 	proxy.OnResponse().DoFunc(func(resp *http.Response, ctx *goproxy.ProxyCtx) *http.Response {
+		logResponse(ctx)
 		if resp == nil && ctx.Error != nil {
 			resp = errorResponse(ctx.Req, ctx.Error)
 		}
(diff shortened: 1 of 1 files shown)


julia-stripe:
@zenazn also i saw you wanted to make smokescreen logs good for debugging in https://github.com/stripe/smokescreen/pull/1 -- would take suggestions for good things to add here.

rb-stripe:
In the spirit of logging all the things, maybe log something about any TLS certs we were served (in the cases we connect to a TLS). common name may be good enough but if you think of something else go for it.

dougbarth-stripe:
LGTM

We should upgrade to the latest goproxy version at some point, but that can be a separate change

evan-stripe:
We should be pretty careful when we do. We've had incidents in the past from accidentally upgrading goproxy where it wouldn't correctly pass through the remote server closing the connection when using CONNECT, or something similar. I don't remember the details, but it's happened a couple times now.

smokescreen.go
```diff
@@ -122,9 +123,12 @@ func buildProxy() *goproxy.ProxyHttpServer {
 			ctx.Resp = errorResponse(ctx.Req, err)
```
rb-stripe: My reading of the code suggests we might never print this? It would be good to know that we're failing lookups but a metric might also be fine.

rb-stripe:
I can't find an unconditional OnComplete hook in goproxy but on the topic of logging both requests and responses it would have been more convenient to log all the info about the request/response on one line (which isn't doable here without shenanigans). This isn't as big an issue now but it will be in at least two scenarios:
1- We're trying to debug mysterious drops of requests (See the other comment about logging the DNS lookup's failure)
2- when we implement access control an early rejection of a request should log something. This is easier to solve than (1) but still a tad annoying.

rb-stripe:
So, since you shipped this I, of course, now have more comments. Sorry. You can ignore me.
- We return early in https://github.com/stripe/smokescreen/blob/master/smokescreen.go#L149 but that means we don't log the reason the request failed. I suspect this came about from my comment about masking resolver failures but maybe we could log the error and the log line or put the error in the log line in some way? (this would make parsing the lines harder so I'm not sure it's a good idea)
- What do you think about adding span-ids? This would let us add the id for an error line and the request log line, to help correlate them. It might be a problem if we haven't open sourced that library. In the long term, this would help us trace outbound proxy requests from their source.

PR #3, opened 2017-01-09
Don't panic when ctx.Resp is nil

julia-stripe:
r? @dougbarth-stripe

For some reason `ctx.Resp` sometimes seems to be nil, and then the logging code panics.

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -139,13 +139,22 @@ func buildProxy() *goproxy.ProxyHttpServer {
 }
 
 func logResponse(ctx *goproxy.ProxyCtx) {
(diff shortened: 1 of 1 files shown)


dougbarth-stripe:
LGTM.

Looks like this happens if the server connection is closed before a response is sent back. I was able to recreate this using `nc -l 8080`.

Steps to reproduce:
1. Run `nc -l 8080` (doing this locally will require temporarily disabling smokescreen's checks for private IPs)
2. Use curl to send a connect to `nc`: `curl --proxy localhost:4750 http://localhost:8080/`
3. Kill the `nc` process after the request is sent

PR #4, opened 2017-01-09
Adds support for PROXY protocol

dougbarth-stripe:
This adds support for the PROXY protocol, the original client's IP to be
tunneled through to smokescreen without editing the original request.
This can be useful when putting a TCP load balancer (eg. a Classic ELB)
in front of the service.

This support needs to be enabled via a new option: `-proxy-protocol`

The underlying library tries to do the right thing by examining the
initial bytes in the request. If the PROXY protocol isn't being used,
the request will still be processed.

The connection's address will be reported based on what the PROXY
protocol passed, so only requests from trusted upstreams should be
allowed (otherwise malicious systems could spoof their IP).

r? @julia-stripe
cc @evan-stripe

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/DataDog/datadog-go/statsd"
+	"github.com/armon/go-proxyproto"
 	"github.com/icub3d/graceful"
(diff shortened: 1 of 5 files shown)


julia-stripe:
lgtm!

PR #5, opened 2017-01-10
Add a graceful shutdown timeout

julia-stripe:
smokescreen was not always shutting down when it got SIGTERM/SIGKILL. This adds a 30 second timeout after which it will shut down.

I switched to a different graceful shutdown library because the old one did not support timeouts.


evan-stripe:
I think it's possible that [goji's graceful package](https://godoc.org/github.com/zenazn/goji/graceful) would serve us better here - it specifically has support for closing idle connections as part of graceful shutdown, which I think could be contributing to the problems we're seeing.

(I don't know offhand who all is connecting to our smokescreen instance but if, e.g., any Go clients were, they'd hold the connection open indefinitely AIUI)

julia-stripe:
@evan-stripe just so I understand -- why is closing idle connections better than having a timeout after which we close all connections?

evan-stripe:
That's a fair question. Thinking about it, I think having a timeout as a backstop means that it doesn't matter, but closing idle connections would probably mean that the old process will finish exiting sooner.

We also in general have a lot more internal usage of goji's graceful (unsurprising, given the author), and almost no usage of tylerb's graceful, so there's some value in consolidation.

julia-stripe:
yep definitely! I'll definitely switch to graceful if only so that we have more homogeneity.

julia-stripe:
@evan-stripe can you take another look at this? I tried to emulate mproxy's graceful listener setup, and in my testing so far (and from my understanding of how the goji code works) this seems to work right.

evan-stripe:
lgtm
on a quick skim, but if you want a detailed review I don't hve time to do one today and am out through Tuesday.

julia-stripe:
@carl-stripe hiii can you look at this to make sure it looks like a reasonable use of goji?

carl-stripe:
I'm not certain this will do what you want. If you look at the documentation for https://godoc.org/github.com/zenazn/goji/graceful#WrapListener, you'll see it calls the underlying `graceful/listener` package in `Deadline` mode (https://godoc.org/github.com/zenazn/goji/graceful/listener#pkg-constants), which matches calls to `Read` (which, if it returns with data, indicates the socket is in use) with calls to `SetReadDeadline` (which is a common thing in several packages—including especially `net/http`—call when they're done processing one request and are ready to read the next one; note that this is a lurktastical hack) to determine if the socket is in-use.

Unfortunately, while `net/http` helpfully calls these functions, `goproxy` Hijacks (https://godoc.org/net/http#Hijacker) connections in order to implement its `CONNECT` proxy. This effectively tells `net/http` to stop managing the socket (so `goproxy` can), and means that `goji/graceful` is left with ~no information about the state of sockets under its management.

I've tried to fix this problem in smokescreen before, and the only way I could think to implement this is to instrument (or find existing instrumentation hooks in) `goproxy` to expose socket states. A less-good MVP solution is to implement the following policy (which is only a few lines, and doesn't require any libraries): `Close` the listening socket, wait ~30 seconds, and exit the program. With this approach you prevent new connections from being established, but you rely on the timeout (and not fine-grained socket tracking) to minimize the amount of damage you have on in-flight requests.

julia-stripe:
@carl-stripe okay, I wrote some extremely simple timeout code like you suggested. what do you think?

smokescreen.go
```diff
@@ -217,17 +214,42 @@ func main() {
...
+	// Runs the server and shuts it down when it receives a signal.
+	//
+	// Why aren't we using goji's graceful shutdown library? Great question!
+	//
+	// There are several things we might want to do when shutting down gracefully:
+	// 1. close the listening socket (so that we don't accept *new* connections)
+	// 2. close *existing* keepalive connections once they become idle
+	//
+	// It is impossible to close existing keepalive connections, because goproxy
+	// hijacks the socket. So all we can do is close the listening socket when
```
carl-stripe: I'd add `because goproxy hijacks the socket and doesn't tell us when they become idle`

smokescreen.go
```diff
@@ -217,17 +214,42 @@ func main() {
 		}
 	}
 
-	server := graceful.NewServer(&http.Server{
+	server := http.Server{
 		Handler: handler,
-	})
+	}
+	timeout := time.Second * time.Duration(60)
```
carl-stripe: I'd probably stick this in a constant at the top (and e.g., let `runServer` look at the constant directly):

`const exitTimeout = 60 * time.Second`

(you don't need to cast 60 to a `time.Duration`—it's an untyped constant)

carl-stripe:
Two inconsequential notes, but
LGTM
now or later.

PR #6, opened 2017-01-11
Log CONNECT proxy requests

julia-stripe:
r? @dougbarth-stripe cc @rb-stripe 

This adds a log line for CONNECT proxy requests. The log lines look like this: (the same format as the other requests we log).

I also added a proxy_type field to the log line.

```
2017/01/11 09:51:33 Received CONNECT request: proxy_type=connect src_host="::1" src_port=55395 host="stripe.com:443" dest_ip="54.187.119.242" dest_port=443 start_time=1484157093 end_time=1484157093 content_length=0
```

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -139,18 +139,27 @@ func buildProxy() *goproxy.ProxyHttpServer {
 	return proxy
 }
 
+func extractHostname(ctx *goproxy.ProxyCtx) string {
+	var hostname string
+	if (ctx.Req != nil) {
+		hostname, _, _ = net.SplitHostPort(ctx.Req.Host);
+	}
+	return hostname
+}
+
 func logHttpsRequest(ctx *goproxy.ProxyCtx, resolved string) {
 	var contentLength int64
 	if ctx.Resp != nil {
 		contentLength = ctx.Resp.ContentLength
 	}
+	hostname := extractHostname(ctx)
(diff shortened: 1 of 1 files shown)


dougbarth-stripe:
Looks like the `host` field will always have a port number in it for `CONNECT` requests, but non-`CONNECT` requests that don't specify a port (and therefore use a default of port 80) won't include it. Seems like we should normalize this data. Since `dest_port` already captures the port, I'd suggest we strip port from the `host` field for both `CONNECT` and non-`CONNECT` connections.

Also, it looks like `content_length` will always be 0. That'll require that we correlate against flow logs to determine how much traffic went out with a particular request. I _think_ if we hijack the `CONNECT` request, we could capture bytes transferred directly. Looking at the [example code](https://github.com/elazarl/goproxy/blob/v1.0/examples/eavesdropper/main.go) that seems a lot more involved though. I'd probably just go with flow logs for now.

julia-stripe:
Okay, I took off the port number! Thoughts now?

rb-stripe:
I've assumed that we want to not hijack CONNECT requests (though it's still annoying that we don't have a hook when the connection terminates that also has the number of bytes transferred). I think it's fine since having the egressboxes not have direct access to the data moved about means one less thing to worry about when securing them.

dougbarth-stripe:
LGTM

PR #7, opened 2017-04-05
Cherry-pick go transport fixes to vendored vendored transport

raylu-stripe:
This fixes hanging CLOSE_WAIT connections when the "Connection: close" header is sent. All 3 commits in https://github.com/golang/go/issues/1967 are cherry-picked here.
https://github.com/golang/go/commit/4900af08391d0ecf566afc7a15b4b4effc5652cb
https://github.com/golang/go/commit/299e7d20a36841828d8a54ec1f4e163e88d460fc
https://github.com/golang/go/commit/1f107b58f7fdac2daf5567673b25a05766b4f29c

transport.go is a file in the go source tree (net/http/transport.go). goproxy vendored transport and we vendored goproxy. Since then, goproxy has deprecated transport (https://github.com/elazarl/goproxy/issues/94).

We can't switch to the new goproxy that uses go's transport because we depend on RoundTripDetails' TCPAddr for logging, a feature removed 3 years ago (see https://github.com/elazarl/goproxy/commit/eb94583 and https://github.com/elazarl/goproxy/commit/456ac5d).

We also can't use the intended replacement (replace ctx.RoundTripper, example in httpdump.go) because that depends on the now-deprecated, vendored transport.go that I am patching which exhibits the bug.

### DIFF
--- a/vendor/github.com/elazarl/goproxy/transport/transport.go
+++ b/vendor/github.com/elazarl/goproxy/transport/transport.go
@@ -511,6 +511,7 @@ type persistConn struct {
 	t        *Transport
 	cacheKey string // its connectMethod.String()
 	conn     net.Conn
+	closed   bool                // whether conn has been closed
 	br       *bufio.Reader       // from conn
 	bw       *bufio.Writer       // to conn
 	reqch    chan requestAndChan // written by roundTrip(); read by readLoop()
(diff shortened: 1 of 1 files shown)


carl-stripe:
Ew.

Can we fork `goproxy`, make our patches, and vendor them in that way? Otherwise someone is going to have a Really Fun Time™ with this regression when they try to update `goproxy` in the future.

Extra bonus points if you shave the `goproxy` yak and make everything less shitty in the process

PR #8, opened 2017-04-05
Fix hanging CLOSE_WAIT and host

raylu-stripe:
This fixes hanging CLOSE_WAIT connections when the "Connection: close" header is sent. The underlying problem was in golang/go#1967, but we were on an old version of goproxy that depended on a vendored version of transport.go that still had the bug.

This also upgrades goproxy to the latest version on github (the version on gopkg.in is from Sept 2014 and the project has commits as recent as Apr 1, 2017).


carl-stripe:
How do you know this works?

raylu-stripe:
(testing discussed elsewhere)

PR #9, opened 2017-05-03
Hide 'X-Smokescreen-Error' header in responses from upstream

adunham-stripe:
This prevents a malicious upstream from being able to forge error responses in things using the proxy.

### DIFF
--- a/smokescreen.go
+++ b/smokescreen.go
@@ -26,6 +26,8 @@ var track *statsd.Client
 
 const exitTimeout = 60 * time.Second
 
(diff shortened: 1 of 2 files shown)


adunham-stripe:
r? @raylu-stripe 
cc @aaron-stripe @karla-stripe

aaron-stripe:
\o/ thank you!

raylu-stripe:
LGTM
I don't think anything consumes this header right now?

adunham-stripe:
I believe that's correct, but worth doing anyway 😄

PR #10, opened 2017-10-02
Migrate from govendor to dep

ChimeraCoder:
Same as when we paired, except I rebased and removed the `_vendor-TIMESTAMP` backup directory that dep creates on initialization, because we don't need it.

govendor doesn't vendor example code or test files, which dep does, so those files are new in the diff. Aside from those, there are only 2-3 changes involving goproxy that *aren't* renames, and they're packages that I don't think we use. I think that's because `proxy_test.go`, which we now pull in, imports `goproxy/ext/image`, which imports `goproxy/regretable`, and so forth.

This also adds testing support for Go 1.9.

Note: Once https://github.com/stripe/goproxy/pull/1 is merged, we'll want to upgrade the version we pin to here.

r? @carleton-stripe 
cc @stripe/observability


Gopkg.toml
```diff
@@ -0,0 +1,40 @@
...
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/stripe/go-einhorn"
+
+[[constraint]]
+  name = "github.com/elazarl/goproxy"
+  source = "git@github.com:stripe/goproxy.git"
+  revision = "92f3b809c67db40d0cecefbf83bb76598614c8f3"
+  #revision = "0890fc15d1b6428a674b8252a2d238ca589daf73"
```
carleton-stripe: Did you intend to leave this one commented out here? Adding a little more context about what that SHA is might be useful

carleton-stripe:
LGTM with optional comment thing

ChimeraCoder:
Self approving after comment + commit-squash.

PR #18, opened 2018-08-29
Fix usage of default ACL with missing role

andrew-stripe:
Currently the proxy always denies if no role is provided. It should instead fall back on the default. This ended up being more test refactoring than real code change as I tried to write a test that actually caught the bug :).

r? @tremblay-stripe 
cc @rlk-stripe


cmd/integration_test.go
```diff
@@ -1,4 +1,4 @@
-// +build integration
```
andrew-stripe: It's very 😢 that these don't run in Travis and by default locally since they represent almost the only test coverage of the core logic here... it looks like this change breaks the tests on Travis though.

tremblay-stripe: I totally agree :/

PR #21, opened 2018-09-05
Log a single detailed line per request

andrew-stripe:
This threads the necessary information around our handlers to log a single line per request with all the details we had been logging. By running everything through a single log method we can mostly guarantee that different request paths don't end up with subtly different logging behavior. I was really hoping this would result in a net decrease in the amount of code but so it goes.

r? @tremblay-stripe

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -19,6 +19,8 @@ import (
 	"testing"
 
 	"github.com/hashicorp/go-cleanhttp"
+	"github.com/sirupsen/logrus"
+	logrustest "github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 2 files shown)


cmd/integration_test.go
```diff
@@ -64,9 +66,35 @@ func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error)
...
+
+	var entries []*logrus.Entry
+	for _, entry := range logs {
+		if entry.Level < logrus.WarnLevel {
+			a.Failf("unexpected log line more severe than Warn", "%v", entry)
+		} else if entry.Level < logrus.DebugLevel {
+			entries = append(entries, entry)
+		}
+	}
+	a.Equalf(1, len(entries), "Expected only a single INFO/WARN log line per request, got: %v", entries)
```
andrew-stripe: Is it going overboard asserting that there's a single line per request? We could relax this to require only a single "proxy_response" line per request.

cmd/integration_test.go
```diff
@@ -64,9 +66,35 @@ func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error)
...
+
+	entry := entries[0]
+	a.Equal(entry.Message, "proxy_response")
+
+	a.Contains(entry.Data, "allow")
```
andrew-stripe: I didn't test all the entries, just a couple key ones. I could check more if you think it's worth it.

cmd/integration_test.go
tremblay-stripe: I'm okay with a single "proxy_response" line per request!

PR #22, opened 2018-09-13
Allow TLS health checks w/o client cert.

rlk-stripe:
Previously, when TLS was configured to request client certs, all requests w/o a client cert would be rejected, including those for `/healthcheck`.  This makes certs optional, allowing health checks to succeed.  

This also means that no-cert requests will fall through to the "default" ACL.  I plan to add an option to prevent this in a future patch.  For now, an implementation could work around this behavior by defining a `RoleFromRequest` that fails with a non-`MissingRoleError`.

This PR also includes some fixes for incorrect type assertions on aliases of `error` (such assertions are always true, as an alias of `error` also fulfills the `error` interface) and other small bugs exposed by the main change.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -406,11 +406,9 @@ func checkIfRequestShouldBeProxied(config *Config, req *http.Request, outboundHo
 	}
 
 	role, roleErr := config.RoleFromRequest(req)
-	if roleErr != nil {
-		// A missing role is OK at this point since we may have a default
-		if _, ok := roleErr.(MissingRoleError); !ok {
-			return nil, roleErr
-		}
+	// A missing role is OK at this point since we may have a default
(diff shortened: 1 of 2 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -40,7 +40,9 @@ type ctxUserData struct {
 	decision *aclDecision
 }
 
-type denyError error
+type denyError struct {
```
andrew-stripe: Why is this form preferred over the other?

rlk-stripe: We're defining `denyError` because we want to later check if a particular `error` is a `denyError`.  But the first one just creates an alias, and because `error` is an interface, *any* `error` is also a `denyError` -- type assertions for interfaces are just duck-typing.

Here's a simplified example: https://play.golang.org/p/X7hny-SZ6a0

PR #23, opened 2018-09-14
Improve handling of --version arg

rlk-stripe:
* Provide an actual SemVer-style version (I'm just using `0.0.1` for now) plus a commit hash (using git's `ident` attribute to translate `$Id$` on checkout).
* Document that NewConfiguration my return `nil` even if no error is provided, and update `--help` handler to do this as well rather than directly exiting.
* Update `main.go` to handle the above case by just exiting cleanly rather than logging a fatal error.

cc @andrew-stripe

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+pkg/smokescreen/constants.go ident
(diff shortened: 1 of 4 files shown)


tremblay-stripe:
LGTM

PR #24, opened 2018-09-24
extract default ranges and localhost from config

rlk-stripe:
Some preliminary work before landing configuration via a file:

- The `--deny-range` configuration (`CidrBlacklist` internally) now only contains user-specified entries; the default private ranges are checked independently.  `allow-range` entries will still take precedence.
- Added IP classification constant to distinguish default vs explicit denied ranges, and renamed for consistency.  Also made the constants private, tweaked the associated log strings, and added statsd-specific strings.
- Removed the `AllowProxyToLoopback` config option; `--allow-range 127.0.0.1/32` (or the IPv6 equivalent) can be used instead.

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -56,7 +56,7 @@ func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error,
 		if !a.NoError(err) {
 			return
 		}
-		a.Equal(200, resp.StatusCode)
+		a.Equal(200, resp.StatusCode, "TestCase was %v", test)
 	} else {
 		if !a.NoError(err) {
 			return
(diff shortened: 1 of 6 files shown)


rlk-stripe:
cc @andrew-stripe

andrew-stripe:
This works for me as long as `allow-range` still takes precedence. I just pass `--allow-range 0.0.0.0/0`

rlk-stripe:
Yep.  For the record, `allow-range` is now top priority, overriding the `IsGlobalUnicast` check in addition to `IsLoopback`, so specifying `0.0.0.0/0` will allow broadcast, link-local, etc. addresses.

andrew-stripe:
Ah ok, I'd probably prefer to maintain the IsGlobalUnicast check to avoid clients doing weird and unexpected things but it's probably fine since I'll generally be using domain whitelists. 

The interleaving of these flags and embedded defaults feels a bit hard to reason about. I feel like I'd rather be able to specify a list of rules that execute in order until something matches, e.g.:
```
IsGlobalUnicast REJECT
127.0.0.1/24 REJECT
10.1.2.3.4/32 ACCEPT
PrivateNetwork REJECT
0.0.0.0/0 ACCEPT
```

May not be worth it right now since neither of us really need it.

rlk-stripe:
> May not be worth it right now since neither of us really need it.

That's pretty much my thinking on the subject.

andrew-stripe:
Damn, nerd-snipe failed :)

tremblay-stripe:
LGTM

PR #25, opened 2018-09-25
simplify cert loading

rlk-stripe:
r? @tremblay-stripe 
cc @andrew-stripe 

(This is based on top of my previous PR, which isn't in master yet, so you may want to just look at the last commit's diff.)

It turns out that`tls.LoadX509KeyPair` will happily extract the relevant PEM blocks from the supplied files, so we can cut out a fair amount of manual parsing code by just passing our bundle file as both of the `certFile` and `keyFile` args.  I've also made SetupTls take two files, pushed the "use the same file twice" logic out a level to `NewConfiguration`.  (I expect a future CLI change could allow an optional key file arg, but it's not implemented yet.)

`NewConfiguration` now only calls `SetupTls` if a file is specified, and `SetupTls` errors if you call it without cert/key files.  (Client CA files remain optional.)

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -56,7 +56,7 @@ func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error,
 		if !a.NoError(err) {
 			return
 		}
-		a.Equal(200, resp.StatusCode)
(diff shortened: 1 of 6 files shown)


tremblay-stripe:
This makes me very happy
LGTM

PR #26, opened 2018-09-25
YAML config (part 1)

rlk-stripe:
Add an `UnmarshalConfig` func (and `LoadConfig` convenience wrapper) which reads YAML configuration data.  Not all config options are supported yet.

Add a `check_config` program that (currently) just loads a config file and dumps the resulting `Config`.

Misc config cleanup: 

- Take advantage of previous allow/deny list handling chages to simplify setup of these values.
- Move default statsd namespace prefix to a global, and split out `SetupStatsdWithNamespace` from `SetupStatsd` so prefix is optional.
- Extract static host matching regex from config struct to a global

No changes to CLI interface in this PR (so you can't actually use config files in the main smokescreen binary yet).

r? @tremblay-stripe 
cc @andrew-stripe


pkg/smokescreen/constants.go
```diff
@@ -17,6 +22,14 @@ var privateNetworkStrings = [...]string{
 
 var PrivateNetworkRanges []net.IPNet
 
+// Using a globally-shared Regexp can impact performace due to lock contention,
+// but calling Copy() for each connection is much worse, and it looks like
+// handing out Regexps from a pool doesn't save us anything either, so we'll
+// just live with it.
```
tremblay-stripe: Considering that the rest of the workload is IO bounded

PR #27, opened 2018-10-02
Rlk/config part 2

rlk-stripe:
r? @tremblay-stripe 

- Pull default RoleFromRequest implementation out of `config.go` / package `smokescreen`.
  - A cert-only version is created in the `main` package for our `smokescreen` executable.
  - The integration test uses both cert and header-based implementations.
- Replaced `Init()` method which needed to be called after setting up other config with `NewConfig`, which returns a pointer to a new Config object with maps and the Log member initialized.
- Extracted logic around calling RoleFromRequest into a `getRole` function.
- Misc cleanup of formatting and integration test output.

### DIFF
--- a/cmd/check_config/main.go
+++ b/cmd/check_config/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"github.com/stripe/smokescreen/pkg/smokescreen"
 )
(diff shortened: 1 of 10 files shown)


cmd/check_config/main.go
```diff
@@ -19,6 +19,13 @@ func main() {
 		os.Exit(1)
 	}
 
-	fmt.Printf("Parsed: %#v\n", config)
+	fmt.Printf("Parsed: %#v\n\n", config)
+
+	if config.Port < 1 || config.Port > 65535 {
```
tremblay-stripe: Should this be checked when loading the config itself?

cmd/check_config/main.go
```diff
@@ -19,6 +19,13 @@ func main() {
...
+		fmt.Printf("Invalid port: %d", config.Port)
+	}
+
+	//TODO more checks here
```
tremblay-stripe: Can we remove this?

rlk-stripe:
Removed the comment, and moved the error-checking logic into config.go, though it's just a helper method, not integrated into `LoadConfig`.  Thinking about a separate `LoadAndCheckConfig` or maybe `SafeLoadConfig` for that.  Thoughts?

rlk-stripe:
I ended up removing `config.Check()` entirely, as the only thing it was checking was the port number.  I changed the type of that to `uint16`, so invalid values aren't possible.  (Zero is technically valid, as a "pick for me" flag, though IDK if that's actually useful.)

I think this is ready go be merged now.

PR #29, opened 2018-10-22
Error handling cleanup

rlk-stripe:
- Consolidate error handling.
  - `checkIfRequestShouldBeProxied` handles errors internally, and always returns a decision with a user-visible reason.
  - This allows us to remove `UnknownRoleError` entirely.
- Normalize error message formatting: capitalization, colons, (mostly) active voice sentences.
- Add some more logging.

Also fixed a crashing bug when trying to use the default `RoleFromRequest` without TLS.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -443,6 +443,11 @@ func getRole(config *Config, req *http.Request) (string, error) {
 	case IsMissingRoleError(err) && config.AllowMissingRole:
 		return "", nil
 	default:
+		config.Log.WithFields(logrus.Fields{
+			"error": err,
+			"is_missing_role": IsMissingRoleError(err),
+			"allow_missing_role": config.AllowMissingRole,
+		}).Info("Unable to get role for request")
 		return "", err
 	}
 }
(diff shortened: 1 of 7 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -440,61 +433,69 @@ func getRole(config *Config, req *http.Request) (string, error) {
...
-	switch result {
+	switch action {
 	case EgressAclDecisionDeny:
-		decision.reason = "role is not allowed to access this host"
-		return decision, nil
+		decision.reason = "Role is not allowed to access this host"
 
 	case EgressAclDecisionAllowAndReport:
-		decision.reason = "role is not allowed to access this host but report_only is true"
+		decision.reason = "Role is not allowed to access this host but report_only is true"
```
tremblay-stripe: Shall we fmt.Sprintf the said role?

tremblay-stripe: It would make it easier to look entries up in short term log

tremblay-stripe: It could also be another PR. (I can get on that)

rlk-stripe: I had some vague FUD about echoing the role, like if a downstream service lets a user control it, and then displays the proxy's response in a web page or something injectable.  Admittedly, smokescreen probably isn't the weakest link there.

rlk-stripe: If you want to do it, I'll approve it.

PR #30, opened 2018-10-31
Simplify categorization

tremblay-stripe:
Remove usage of `net` library in order to determine address type
Rely on provided default DenyRanges


rlk-stripe:
I have some concerns about bypassing the `IsGlobalUnicast` check here.  For example, the default deny list in this PR misses the zero-valued IPv6 address (`[::]` or `[::0]`, which effectively exposes services listening on localhost.)

PR #33, opened 2018-11-07
constants: Add link-local address ranges

dubek:
Add link-local address ranges (IPv4 and IPv6) to the privateNetworkStrings constant.

The list is based on https://en.wikipedia.org/wiki/Link-local_address .

A possible vulnerable scenario would be a Smokescreen proxy running on EC2 machine; the end-user (attacker) can then ask the proxy to connect to `http://169.254.169.254/latest/meta-data` which will return Amazon EC2 internal information; details in https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html . With the proposed fix this request will be blocked.

### DIFF
--- a/pkg/smokescreen/constants.go
+++ b/pkg/smokescreen/constants.go
@@ -15,9 +15,11 @@ const DefaultStatsdNamespace = "smokescreen."
 
 var privateNetworkStrings = [...]string{
 	"10.0.0.0/8",
+	"169.254.0.0/16",  // Link-local address range
 	"172.16.0.0/12",
 	"192.168.0.0/16",
 	"fc00::/7",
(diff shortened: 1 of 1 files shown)


tremblay-stripe:
Hi @dubek and thank you very much for this PR!

The link local case is already covered in [1][2] the ip classification code. The need for `PrivateNetworkRanges` comes from private network blocks also being considered as global unicast.

Thank you for contributing!

[1] https://github.com/stripe/smokescreen/blob/61fb5eb5382bd1178999ec1de20b3fa964c0f431/pkg/smokescreen/smokescreen.go#L100
[2] https://golang.org/src/net/ip.go#L165

PR #39, opened 2019-01-16
Rule resolution

tremblay-stripe:
Not finding a rule is not an application error
Usage of default rule is now a tag in ACL decision
r?: @rlk-stripe

### DIFF
--- a/pkg/smokescreen/acl_loader_v1.go
+++ b/pkg/smokescreen/acl_loader_v1.go
@@ -27,7 +27,7 @@ func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclD
 
 	rule := ew.ruleForService(fromService)
 	if rule == nil {
-		action = EgressAclDecisionNoRuleDeny
+		action = EgressAclDecisionDeny
 		return action, false, nil
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/smokescreen/acl_loader_v1.go
```diff
@@ -22,38 +22,41 @@ type EgressAclConfig struct {
 	Default  *EgressAclRule
 }
 
-func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclDecision, error) {
+func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclDecision, bool, error) {
+	var action EgressAclDecision
+
 	rule := ew.ruleForService(fromService)
 	if rule == nil {
-		return 0, fmt.Errorf("unknown role: '%s'", fromService)
```
rlk-stripe: `ruleForService` returns nil only if no default rule is configured.  Do we care about distinguishing this case from the default rule being explicitly set to "deny"?  I'm generally not thrilled with adding a special case to the enum and also another return parameter.

tremblay-stripe: Agreed!

tremblay-stripe:
ptal @rlk-stripe

PR #45, opened 2019-03-28
allow other processes to know about current connection statuses

tremblay-stripe:
Smokescreen will open an UDS which will allow other processes (by default, from the same user or group) to know about the state of established connections.

Sample output:
```
➜  smokescreen git:(tremblay/conn-tracking) ✗ curl -vvvv --unix-socket ./track/track-19049.sock -s http://stats/ | jq .
*   Trying ./track/track-19049.sock...
* Connected to stats (track/track-19049.sock) port 80 (#0)
> GET / HTTP/1.1
> Host: stats
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 28 Mar 2019 22:46:51 GMT
< Content-Length: 190
< 
{ [190 bytes data]
* Connection #0 to host stats left intact
[
  {
    "id": "824634441944",
    "role": "",
    "rhost": "google.com:443",
    "created": "2019-03-28T15:46:31.307748-07:00",
    "bytesIn": 196608,
    "bytesOut": 520,
    "wakeups": 10,
    "secondsSinceLastActivity": 20.112740421
  }
]
```

### DIFF
--- a/pkg/smokescreen/stats_server.go
+++ b/pkg/smokescreen/stats_server.go
@@ -2,23 +2,22 @@ package smokescreen
 
 import (
 	"fmt"
-	"net/http"
 	"net"
+	"net/http"
 	"os"
 )
 
-
 type StatsServer struct {
-	config *Config
-	ln net.Listener
-	mux *http.ServeMux
+	config     *Config
+	ln         net.Listener
+	mux        *http.ServeMux
 	socketPath string
 }
 
 func newServer(config *Config) (s *StatsServer) {
 	s = &StatsServer{
 		config: config,
-		mux: http.NewServeMux(),
+		mux:    http.NewServeMux(),
 	}
 
 	s.mux.HandleFunc("/", s.stats)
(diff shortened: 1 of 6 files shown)


pkg/smokescreen/stats_server.go
```diff
@@ -0,0 +1,84 @@
...
+	return
+}
+
+func (s *StatsServer) Serve() {
+	pid := os.Getpid()
+	s.socketPath = fmt.Sprintf("%s/track-%d.sock", s.config.StatsSocketDir, pid)
+	ln, err := net.Listen("unix", s.socketPath)
+
+	if err != nil {
+		s.config.Log.Fatal("Could not start the reporting server.")
```
rlk-stripe: I think we should include `err` in the message here.

tremblay-stripe: true!

tremblay-stripe: ```{"level":"fatal","msg":"Could not start the reporting server.listen unix track/track-52727.sock: bind: no such file or directory","time":"2019-03-29T10:55:12-07:00"}```

tremblay-stripe:
ptal @rlk-stripe

rlk-stripe:
LGTM

PR #46, opened 2019-04-03
Wait until all connections are idle before exiting

marcdoan:
So I tested this on my laptop by running:
`go run main.go --stats-socket-dir=track`
and then kicked two "long" curl commands a few seconds apart:
`curl --limit-rate 160000 -vvvvx http://localhost:4750 https://raw.githubusercontent.com/dwyl/english-words/master/words.txt`
I then sent a SIGTERM to the process using `kill -15`

I think this should work with deploys if we restart smokescreen-srv with `--stats-socket-dir` and pass in the path to the current deploy _with the symlinks resolved_. Then each deploy will have it's own "track" directory. In the case where the process panics and is restarted by einhorn, each process still has its own UDS file and knows which one to communicate with.

For AWS instance lifecycle restarts, I will just make the terminator script send it a SIGTERM and wait until the process gracefully shuts down.

Still have to add alerting, and test actual deploys but can you guys take a look and see if this on the right track? r? @tremblay-stripe @rlk-stripe

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -44,7 +44,7 @@ type Config struct {
 	StatsServer                  interface{} // StatsServer
 	ConnTracker                  *sync.Map   // The zero Map is empty and ready to use
 	IdleThresholdSec             time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
-	WgCxns                       sync.WaitGroup // This wait group tracks *open* (active & idle) connections for graceful shutdown.
+	WgCxns                       *sync.WaitGroup // This wait group tracks *open* (active & idle) connections for graceful shutdown.
 }
 
 type missingRoleError struct {
@@ -99,7 +99,7 @@ func NewConfig() *Config {
 		ExitTimeout:             500 * time.Minute,
 		StatsSocketFileMode:     os.FileMode(0700),
 		IdleThresholdSec:        10 * time.Second,
-		WgCxns:                  sync.WaitGroup{},
+               WgCxns:                  &sync.WaitGroup{},
 	}
 }
 
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -44,7 +44,7 @@ type Config struct {
 	StatsServer                  interface{} // StatsServer
 	ConnTracker                  *sync.Map   // The zero Map is empty and ready to use
 	IdleThresholdSec             time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
-	WgCxns						 sync.WaitGroup
+	WgCxns                       sync.WaitGroup // This wait group tracks *open* (active & idle) connections for graceful shutdown.
 }
 
 type missingRoleError struct {
@@ -99,7 +99,7 @@ func NewConfig() *Config {
 		ExitTimeout:             500 * time.Minute,
 		StatsSocketFileMode:     os.FileMode(0700),
 		IdleThresholdSec:        10 * time.Second,
-		WgCxns:					 sync.WaitGroup{},
+		WgCxns:                  sync.WaitGroup{},
 	}
 }
 
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -43,8 +43,8 @@ type Config struct {
 	StatsSocketFileMode          os.FileMode
 	StatsServer                  interface{} // StatsServer
 	ConnTracker                  *sync.Map   // The zero Map is empty and ready to use
-	IdleThresholdSec			 time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
-	WaitForAllIdleSec			 time.Duration // Wait this many seconds before checking (again) to ensure all connections are idle before exiting.
+	IdleThresholdSec             time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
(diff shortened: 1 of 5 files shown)


pkg/smokescreen/config.go
```diff
@@ -94,8 +96,10 @@ func NewConfig() *Config {
 		ConnTracker:             new(sync.Map),
 		Log:                     log.New(),
 		Port:                    4750,
-		ExitTimeout:             60 * time.Second,
+		ExitTimeout:             500 * time.Minute,
```
marcdoan: 500 minutes, or 8.33 hours, is the default amount of time you can delay before the termination cycle is completed on AWS.

pkg/smokescreen/smokescreen.go
```diff
@@ -433,11 +433,25 @@ func runServer(config *Config, server *http.Server, listener net.Listener, quit
...
+		// the program has exited normally, wait for all connections to become idle before
+		// continuing in an attempt to shutdown gracefully
+		beginTs := time.Now()
+		alerted := false
+		for {
+			if config.StatsServer.(*StatsServer).GetNumActiveConn() > 0 {
+				config.Log.Info(fmt.Sprintf("There are still active connections. Waiting %v before checking again.", config.WaitForAllIdleSec))
+				if time.Now().Sub(beginTs) > config.ExitTimeout && !alerted {
+					config.Log.Info(fmt.Sprintf("We've been waiting for %v to shut down but there are still active connections.", config.ExitTimeout))
+					// Do alerting here
```
marcdoan: I want to send a Slack alert or Sentry alert here...is that currently possible in this repo?

rlk-stripe: Not currently.  You could do something similar to `Config.RoleFromRequest` where there's a default behavior wired up in Smokescreen's `main()`, but another program that pulls in the `smokescreen` package can provide environment-specific behavior.  (Alternately, you could send a metric to statsd, and trigger an alert on the far side of that?)

pkg/smokescreen/stats_server.go
```diff
@@ -47,6 +48,18 @@ func (s *StatsServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	s.mux.ServeHTTP(w, req)
 }
 
+func (s *StatsServer) GetNumActiveConn() int {
+	n := 0
+	s.config.ConnTracker.Range(func(k, v interface{}) bool {
+		c := k.(*ConnExt)
+		if time.Now().Sub(c.LastActivity) < s.config.IdleThresholdSec {
```
tremblay-stripe: The stats lock on the instrumented connection (ConnExt) should be grabbed before accessing fields in it. ⚠️

tremblay-stripe:
Looking great!
May I interest you to look at `time.Ticker`?
Using a `Ticker` you could avoid sleeping on the thresholds and instead, have a piece of code aware of connection `.Close()`. 

You could also do simpler with a `sync.WaitGroup`. Basically, every time a new connection successfully opens, `wg.Add(1)`. Anytime one closes, `wg.Done()`. And then, when you get the termination signal, you quit on either `wg.Wait()` unblocking or the threshold passing.

Those ways, if all your connections end before delta threshold time, the proxy can quit. 

Let me know if you want to discuss those further! You're doing great :D

pkg/smokescreen/config.go
```diff
@@ -43,6 +43,8 @@ type Config struct {
 	StatsSocketFileMode          os.FileMode
 	StatsServer                  interface{} // StatsServer
 	ConnTracker                  *sync.Map   // The zero Map is empty and ready to use
+	IdleThresholdSec             time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
+	WaitForAllIdleSec            time.Duration // Wait this many seconds before checking (again) to ensure all connections are idle before exiting.
```
tremblay-stripe: I think you won't need `WaitForAllIdleSec` if you implement one of https://github.com/stripe/smokescreen/pull/46#issuecomment-479958041

marcdoan:
re-r? @tremblay-stripe

pkg/smokescreen/smokescreen.go
```diff
@@ -433,12 +433,52 @@ func runServer(config *Config, server *http.Server, listener net.Listener, quit
...
 	}
 
-	if semiGraceful {
-		// the program has exited normally, wait 60s in an attempt to shutdown
-		// semi-gracefully
-		config.Log.WithField("delay", config.ExitTimeout).Info("Waiting before shutting down")
-		time.Sleep(config.ExitTimeout)
-	}
+	if graceful {
+		// The program has exited normally, so wait for all connections to close or become idle before
```
tremblay-stripe: The program hasn't really exited at that point

pkg/smokescreen/smokescreen.go
```diff
@@ -433,12 +433,52 @@ func runServer(config *Config, server *http.Server, listener net.Listener, quit
...
+					break
+				}
+			}
+		}()
+
+		// Wait for the exit signal.
+		<- exit
+	}
+
+	// Close all open (and idle) connections to send their metrics to Splunk.
```
tremblay-stripe: ~Splunk~ -> log

marcdoan:
re-r? @tremblay-stripe 
I tested the following cases:
1. All connections _are already idle_ - shuts down immediately
2. All connections _become idle_ - shuts down after idle check
3. All connections _become closed_ - shuts down after last connection closes
4. Some connections close, and the rest become idle - shuts down after idle check

PR #47, opened 2019-04-08
Update goproxy to pick up fixes to https_proxy config

andrew-stripe:
Picks up https://github.com/stripe/goproxy/pull/5

r? @rlk-stripe


andrew-stripe:
Note that this pulls in a ton of vendor code, especially from `golang.org/x/text/`. I'm not sure we really care.

rlk-stripe:
Yeah, that's a lot of dependencies for basically just distinguishing between "http" and "https", but I don't think it's worth the trouble of trying to avoid them.

LGTM

PR #48, opened 2019-04-09
Add {allow,deny}-address options

andrew-stripe:
This adds `--allow-address` and `--deny-address` options that operate similarly to the range options but allow for specifying a particular port to allow/block. My immediate use for this is allowing the proxy to chain to another proxy running on localhost while preventing it from accessing other services on localhost.

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


pkg/smokescreen/config.go
```diff
@@ -13,18 +13,24 @@ import (
...
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
 	"github.com/DataDog/datadog-go/statsd"
 	log "github.com/sirupsen/logrus"
 )
 
+type RuleRange struct {
```
andrew-stripe: I think `RuleRange` is kind of a weak name. Open to ideas.

rlk-stripe: `RangeWithPort` maybe?  I'm pretty ok with `RuleRange` though.

rlk-stripe:
LGTM

pkg/smokescreen/config.go
andrew-stripe: I can leave as-is if you think it's OK. Sometimes I over-think the naming...

PR #49, opened 2019-04-09
Pass host through to upstream proxy

andrew-stripe:
Currently, when smokescreen handles a CONNECT request to proxy HTTPS it resolves the IP address in the HandleConnectFunc handler and changes the target hostname to the resolved IP address. This is done to ensure that the request only goes to an IP address we've decided is safe. I believe this is unnecessary since goproxy uses our special dialer to make the request which will perform the same validation on the IP address:
https://github.com/stripe/goproxy/blob/806bae1a599998b57e7ba45b89de7750b320df90/https.go#L109-L113

Performing resolution in the HandleConnectFunc handler has a secondary benefit of allowing us to provide a cleaner error message to the client and better logging than we'd otherwise get from a dialer error. Note that because we only pre-resolve for CONNECT requests, we do not get this nicer error messaging for HTTP proxying. 

The disadvantage of pre-resolving and replacing the hostname with an IP address is that if we chain this proxy onto another upstream proxy, the upstream does not see the original hostname. This PR aims to solve that problem while improving our error messaging and logging. It performs pre-resolution for both plain HTTP and CONNECT proxy requests without replacing the host and stores the result in the context we thread through to the dialer. In the dialer, we use the pre-resolved address to avoid a second lookup unless the target address has changed (as it does when we're chaining to another proxy). If the target address has changed, we call `safeResolve` again to ensure we only ever connect to safe addresses.


andrew-stripe:
r? @adunham-stripe 
cc @rlk-stripe

PR #50, opened 2019-04-10
Bump goproxy to fix handling of https_proxy env var

andrew-stripe:
Like it says on the tin. Picks up https://github.com/stripe/goproxy/pull/6

r? @rlk-stripe


rlk-stripe:
r+

PR #52, opened 2019-04-18
Ajj allow deny lists

ajj-stripe:
Added global allow and deny list functionality to smokescreen.

Information about the functionality can be found in the repo's modified README.md

Added automated tests to ensure correct functionality.

This is motivated by a downstream Stripe service wanting this functionality. The most convenient place to put this behavior is in smokescreen, but happy to re-implement in the downstream service if necessary. Would prefer Slack to discuss details to prevent information leak to public Github.


README.md
```diff
@@ -132,6 +132,17 @@ A host can be specified with or without a globbing prefix
 
 [Here](https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/testdata/sample_config.yaml) is a sample ACL.
 
+
+#### Global Allow/Deny Lists
+Optionally, you may specify a global allow list and a global deny list in your ACL config.
+This overrides the settings for any individual ACL entry, including the action.
+For example, specifying `example.com` in your global_allow_list will allow traffic for that domain on a host, even if that host is set to `enforce` and does not specify `example.com` in its allowed domains.
+Similarly, specifying `malicious.com` in your global_deny_list will deny traffic for that domain on a host, even if that host is set to `report` or `open`, regardless of whether or not that host specifies `malicious.com` in its allowed domains.
```
rlk-stripe: It seems like this doesn't line up with the code below.  From my reading, a host-specific `open` will trump a global `enforce`.

pkg/smokescreen/acl_loader_v1.go
```diff
@@ -44,21 +46,44 @@ func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclD
...
-				return EgressAclDecisionAllow, defaultRuleUsed, nil
-			}
-		} else {
-			if host == toHost {
-				return EgressAclDecisionAllow, defaultRuleUsed, nil
-			}
+	// if the host matches any of the rule's allowed domains, allow
+	for _, domainGlob := range rule.DomainGlob {
+		if hostMatchesGlob(toHost, domainGlob) {
+			return EgressAclDecisionAllow, defaultRuleUsed, nil
```
rlk-stripe: If we match and return here, the deny list is never checked.

ajj-stripe:
r? @rlk-stripe 

good catch, thanks!
added a test to prevent regression

rlk-stripe:
Thanks!

LGTM

PR #54, opened 2019-04-18
add project and domain tags for decision metrics

ajj-stripe:
Additional tagging for StatsD metrics

Useful for alerting on specific projects and diagnosing erroneous blocks based on domain


rlk-stripe:
Code change seems fine, but it's just project, vs the PR mentions domain?

ajj-stripe:
Oh, weird, somehow I must have left the rest of the file uncommitted, my bad

ajj-stripe:
Not sure that `destination` is the correct value to place there, please confirm

rlk-stripe:
`destination` is the hostname without the port, which I think makes sense here.

PR #55, opened 2019-04-19
modified global list behavior slightly

ajj-stripe:
My bad. We want to preserve the ability for specific hosts to override the `global_deny_list`.

I doubt that anyone has consumed the previous behavior in the last few hours, so hopefully this change is okay.

I took this opportunity to also specify what happens in the edge case of conflicting global lists.

Of course, added tests to confirm the new expected behavior.

Sorry, 100% my fault I didn't think rigorously through the desired behavior before submitting the previous PR.

r? @rlk-stripe

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,9 +135,15 @@ A host can be specified with or without a globbing prefix
 
 #### Global Allow/Deny Lists
 Optionally, you may specify a global allow list and a global deny list in your ACL config.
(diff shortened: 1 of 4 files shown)


rlk-stripe:
TBH, I'm not sure what our use case for the global deny list would be, but I think if we did use it, we would want specific overrides to be possible, so I'm on board with this change.

rlk-stripe:
LGTM

PR #56, opened 2019-04-19
added logging info from TLS cert

ajj-stripe:
CName info is needed to determine the specific host in the Stripe use case.
OU info is provided to be useful in the default use case (where CName provides the role)


ajj-stripe:
@r? @rlk-stripe

PR #57, opened 2019-04-23
log error on deny when present

ajj-stripe:
Currently, if a request is blocked due to ip ranges, an error is supplied but it is not logged due to `logProxy` logic. The decision reason erroneously presents the request as if it should have passed.

In the short term, any time traffic is blocked and an error is present, we can log the error to aid in diagnosis. In the long term, we may want to refactor the logic for populating `decision_reason` to become more accurate.

This PR should preserve the log level previously. Blocked traffic will warn, and the log will only be an error line if the error is an unexpected (non-`denyError`) type.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -279,10 +279,6 @@ func logProxy(
 		"content_length": contentLength,
 	}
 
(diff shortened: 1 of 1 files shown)


rlk-stripe:
LGTM

Thanks for fixing!

PR #58, opened 2019-04-24
sheng/propagate decision reason

sheng-stripe:
The problem here is that the classification for an action's decision isn't propagated properly, and so we're getting some incorrect decision_reasons logged out to splunk. Tests all pass, but I'm not sure if I've covered all the corner cases.

r? @ajj-stripe @tremblay-stripe


pkg/smokescreen/egressacl.go
```diff
@@ -1,6 +1,6 @@
 package smokescreen
 
 type EgressAcl interface {
-	Decide(fromService string, toHost string) (EgressAclDecision, bool, error)
+	Decide(fromService string, toHost string) (EgressAclDecision, string, bool, error)
```
ajj-stripe: I think it's worth documenting here and in the implementation function header what each of the return values represent.

pkg/smokescreen/smokescreen.go
```diff
@@ -123,21 +123,21 @@ func classifyAddr(config *Config, addr *net.TCPAddr) ipType {
...
 
 	classification := classifyAddr(config, resolved)
 	config.StatsdClient.Incr(classification.statsdString(), []string{}, 1)
 
 	if classification.IsAllowed() {
-		return resolved, nil
+		return resolved, classification.String(), nil
 	}
-	return nil, denyError{fmt.Errorf("The destination address (%s) was denied by rule '%s'", resolved.IP, classification)}
+	return nil, "", denyError{fmt.Errorf("The destination address (%s) was denied by rule '%s'", resolved.IP, classification)}
```
ajj-stripe: If `!classification.IsAllowed()`, the egress will be blocked. In that case, we want to populate the decision reason with something similar to the error message supplied here, right? Because we want to be able to say something like "Denied because resolved IP matched rule X"

pkg/smokescreen/smokescreen.go
```diff
@@ -123,21 +123,21 @@ func classifyAddr(config *Config, addr *net.TCPAddr) ipType {
...
-		return nil, err
+		return nil, "", err
 	}
 
 	classification := classifyAddr(config, resolved)
 	config.StatsdClient.Incr(classification.statsdString(), []string{}, 1)
 
 	if classification.IsAllowed() {
-		return resolved, nil
+		return resolved, classification.String(), nil
```
sheng-stripe: String is overloaded for classification, which is an enum. It provides useful information informing why an IP was allowed or denied which we should propagate up.

pkg/smokescreen/acl_loader_v1.go
```diff
@@ -18,41 +18,41 @@ type EgressAclRule struct {
...
 
-func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclDecision, bool, error) {
+func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclDecision, string, bool, error) {
 	var action EgressAclDecision
 
 	rule := ew.ruleForService(fromService)
 	if rule == nil {
 		action = EgressAclDecisionDeny
-		return action, false, nil
+		return action, "", false, nil
```
ajj-stripe: Worth supplying a reason here, right? "No rule matched" or similar?

pkg/smokescreen/smokescreen.go
```diff
@@ -577,20 +577,20 @@ func checkIfRequestShouldBeProxied(config *Config, req *http.Request, outboundHo
 
 	switch action {
 	case EgressAclDecisionDeny:
-		decision.reason = "Role is not allowed to access this host"
```
sheng-stripe: Previously, this would in certain cases this decision reason would be incorrect, which it should be something like ""Deny: User Configured".

pkg/smokescreen/acl_loader_v1.go
```diff
@@ -18,41 +18,41 @@ type EgressAclRule struct {
...
 	}
 
 	defaultRuleUsed := rule == ew.Default
 
 	// if the host matches any of the rule's allowed domains, allow
 	for _, domainGlob := range rule.DomainGlob {
 		if hostMatchesGlob(toHost, domainGlob) {
-			return EgressAclDecisionAllow, defaultRuleUsed, nil
+			return EgressAclDecisionAllow, "if the host matches any of the rule's allowed domains, allow", defaultRuleUsed, nil
```
ajj-stripe: I think the phrasing for these should resemble more in the active past tense, like "Destination matched a rule in the role's list of allowed_domains"

(same for the next few reason strings)

pkg/smokescreen/smokescreen.go
```diff
@@ -356,7 +356,7 @@ func handleConnect(config *Config, ctx *goproxy.ProxyCtx) (*net.TCPAddr, error)
 		return nil, denyError{errors.New(decision.reason)}
 	}
 
-	resolved, err = safeResolve(config, "tcp", ctx.Req.Host)
+	resolved, decision.reason, err = safeResolve(config, "tcp", ctx.Req.Host)
```
sheng-stripe: My assumption is that should mutate the decision object in go, but uhh this is my first time writing golang. D:

pkg/smokescreen/acl_loader_v1.go
```diff
@@ -62,16 +62,16 @@ func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclD
 	case ConfigEnforcementPolicyEnforce:
 		action = EgressAclDecisionDeny
 	case ConfigEnforcementPolicyOpen:
-		return EgressAclDecisionAllow, defaultRuleUsed, nil
+		return EgressAclDecisionAllow, "", defaultRuleUsed, nil
```
ajj-stripe: We can supply a reason here, like "Role has open enforcement policy"

pkg/smokescreen/acl_loader_v1.go
```diff
@@ -62,16 +62,16 @@ func (ew *EgressAclConfig) Decide(fromService string, toHost string) (EgressAclD
...
 	default:
-		return 0, defaultRuleUsed, fmt.Errorf("unexpected policy value for (%s -> %s): %d", fromService, toHost, rule.Policy)
+		return 0, "", defaultRuleUsed, fmt.Errorf("unexpected policy value for (%s -> %s): %d", fromService, toHost, rule.Policy)
 	}
 
 	// use the decision from rule.Policy
-	return action, defaultRuleUsed, nil
+	return action, "", defaultRuleUsed, nil
```
ajj-stripe: On line 62 we can also assign a reason, and return it alongside the action that was selected here. That way when we hit this case, we can log which policy was selected, and that the domain fell through to the policy action.

tremblay-stripe:
Hi, I'm pretty busy. I defer to @ajj-stripe's judgement. You could also ask @mdoan-stripe. I'm still happy to help in second line!

pkg/smokescreen/acl_loader_v1_test.go
```diff
@@ -190,7 +190,7 @@ func TestLoadFromYaml(t *testing.T) {
 		a.Equal(0, len(acl.GlobalAllowList))
 	}
 
-    // Load a sane config with global lists
+	// Load a sane config with global lists
```
ajj-stripe: We should add tests in this file to ensure the correct reason string is supplied for each case.
- Blocked because it matches an ip rule
- Blocked due to enforce mode
- Blocked due to global deny list
- Allowed due to `allowed_domain` presence
- Allowed due to `open` or `report` mode
- Allowed due to `global_allow_list`

pkg/smokescreen/smokescreen.go
ajj-stripe: Yeah, this should do what you expect under my understanding of golang.

pkg/smokescreen/acl_loader_v1_test.go
sheng-stripe: Thanks will do. Wanted to throw this up to see if I had the right idea.

pkg/smokescreen/acl_loader_v1.go
sheng-stripe: sgtm, will do.

pkg/smokescreen/egressacl.go
sheng-stripe: Sure.

pkg/smokescreen/smokescreen.go
sheng-stripe: Ah, it was my assumption that we'll log out errors when they occur. Is that wrong?

Seems like currently blank strings get set for reason when an error is detected.

pkg/smokescreen/smokescreen.go
sheng-stripe: Cool, thanks.

pkg/smokescreen/smokescreen.go
ajj-stripe: We do log out errors, but it's still worth making the `decision_reason` accurate.
I don't think `decision_reason` is cleared when an error is present, but maybe it is. In that case, we should maybe supply `error encountered, see error field` or something as the `decision_reason`.

I think it's probably a blank string because in the cases where we supply errors, we're not in the correct context to modify/provide a `decision_reason`.

pkg/smokescreen/acl_loader_v1_test.go
sheng-stripe: I'll add some more tests for ip rules.

sheng-stripe: Eh, actually. Not sure if there's an easy way to mock this.

sheng-stripe:
@ajj-stripe re-r?

ajj-stripe:
lgtm, would like to get a more smokescreen-knowledgeable review from rlk

r? @rlk-stripe

pkg/smokescreen/smokescreen.go
```diff
@@ -148,7 +148,7 @@ func dial(config *Config, network, addr string, userdata interface{}) (net.Conn,
 		outboundHost = v.decision.outboundHost
 	}
 
-	resolved, err := safeResolve(config, network, addr)
+	resolved, _, err := safeResolve(config, network, addr)
```
rlk-stripe: I think we want to update the decision reason here, similar to line 359.  (This code path is hit in the case where we're proxying plain HTTP rather than HTTPS or other traffic via CONNECT.)

sheng-stripe:
@rlk-stripe re-r? Thanks!

PR #59, opened 2019-04-25
Revert "add project and domain tags for decision metrics"

hans-stripe:
Reverts stripe/smokescreen#54.

The metric cardinality is too high for our system, so we're removing these metrics.


gabe-stripe:
lgtm

PR #60, opened 2019-04-26
added project tag after metric tag revert

ajj-stripe:
Adds back the `project` metric tag that was removed as part of this revert: https://github.com/stripe/smokescreen/pull/59

The cardinality of this tag is <= the existing `role` tag.

`dest_domain` was the primary offender for metric cardinality, so this addition shouldn't be problematic.

We want the `project` tag so that we can have team-specific alerts any time egress is blocked via an egress proxy.


ajj-stripe:
This PR page should be scrubbed.

PR #62, opened 2019-05-01
Fix bug in handling HTTP requests without a port in the host header

andrew-stripe:
In #49 we started calling `safeResolve` on `req.Host` which in turn calls `SplitHostPort`. HTTP requests do not necessarily include a port in their host header since they default to port 80. In that case, SplitHostPort would return an error because it requires the port.

The primary fix here is only a few lines and is in the last commit. I confirmed that modifying `req.Host` does not change the host header passed to the upstream service by making a manual request against httpbin. I don't *think* this matters but I could imagine a poorly written service not handling it properly.

The remaining commits are described in their commit messages.

r? @rlk-stripe


andrew-stripe:
Oh, I also confirmed that a test fails after ensuring we always run the safeResolve logic, even without ACLs.

rlk-stripe:
LGTM

PR #63, opened 2019-05-02
Fix logging of dialer errors to retain original 'allow'

andrew-stripe:
#49 caused us to set `decision.allow = false` for all errors from `safeResolve`. That makes it hard to distinguish people passing invalid hostnames from changes in decision behavior. I believe we should instead treat this as an error and retain the original `allow` state on errors. Making this work involved three parts:
* Threading non-deny lookup errors back from checkIfRequestShouldBeProxied
* Correctly threading errors through to the response handler in the http case
* Only setting `allow` in CANONICAL-PROXY-LINE on the basis of the decision, not the error.

I'm not 100% sure the change to allow logging is right but it seems more reasonable to me. What do you think?

r? @rlk-stripe

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -293,8 +293,6 @@ func logProxy(
 
 	fromHost, fromPort, _ := net.SplitHostPort(ctx.Req.RemoteAddr)
 
-	allow := false
-
 	fields := logrus.Fields{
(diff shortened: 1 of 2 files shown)


andrew-stripe:
Approving after rebase

andrew-stripe:
cc @sheng-stripe since it looks like you're also looking at logging

PR #64, opened 2019-05-29
changed service unavailable message

ajj-stripe:
Changes the status message from the default `"Service Unavailable"` to something more specific to smokescreen functionality to help aid in diagnosing unexpected rejection issues.


ajj-stripe:
r? @rlk-stripe 

Having trouble with good messaging that is `smokescreen` use-case agnostic, any suggestions?

PR #66, opened 2019-06-12
avoid mutating host header and respect URL scheme

cds2-stripe:
#62 added a check to ensure there was a `net.ResolveTCPAddr` parsable address in `req.Host`, as this field was used in `safeResolve` to split out the remote host address and port. This caused issues because it was actually mutating the host header passed to upstream services, and was also appending port 80 to non-Connect HTTPS proxy requests.

I also believe httpbin may actually be returning the host headers incorrectly, as I am getting different responses from Go's httptest.Server and httpbin. httptest.Server shows our mutated version of `req.Host` set as the host header.

This PR avoids mutating `req.Host` and moves it to a scheme-respecting variable which is passed along with the request to `safeResolve`.

https://jira.corp.stripe.com/browse/PRIVACYENG-1475?

r? @andrew-stripe
cc @edlee-stripe 
cc @rlk-stripe  
cc @abull-stripe

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -12,7 +12,7 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/armon/go-proxyproto"
+	proxyproto "github.com/armon/go-proxyproto"
 	"github.com/elazarl/goproxy"
 	"github.com/sirupsen/logrus"
 	"github.com/stripe/go-einhorn/einhorn"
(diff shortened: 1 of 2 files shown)


andrew-stripe:
Gah, good catch and fix. I was relying on httpbin returning the correct headers in testing. Can you add a test for this that uses httptest.Server instead of httpbin? Probably would go in https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/smokescreen_test.go

r? @cds2-stripe

rlk-stripe:
Hmm... we can use `httptest` to test a case with an explicit port, but not the default port cases.

I manually tested the plain HTTP default port case for this branch with something like 

```
sudo nc -l 80
./smokescreen --listen-ip 127.0.0.1 --allow-range 127.0.0.1/32
curl -x http://localhost:4750 http://localhost
# watch nc output for "Host: locahost" rather than "Host: localhost:80"
```

cds2-stripe:
@rlk-stripe beat me to it. It isn't possible to write tests for the default cases due to the privileges required to listen onto those ports (80/443). My last commit adds a unit test for checking against explicit ports.

cds2-stripe:
Also verified locally. 

re-r? @andrew-stripe

andrew-stripe:
Ah, you're right and I forgot that I ran into the same problem trying to test my own PR (which is I guess why this cropped up in the first place...).

:+1:

PR #67, opened 2019-06-24
smokescreen healthcheck improvements

cds2-stripe:
This PR moves the `/healthcheck` endpoint behavior out of smokescreen and into a user defined `http.Handler` specified in the configuration and creates an atomic `IsShuttingDown` indicator which healthchecks can use to understand if smokescreen is actively shutting down.

The previous `/healthcheck` implementation was tightly coupled to to the way stripe-health works today, and by making this a generic `http.Handler` controlled by the user we can move the healthcheck implementation into egress-proxy-srv. Living inside gocode means we can use the `core/healthcheck` libraries.

`IsShuttingDown` was as a fix for a previous incident. There are more details are in PRIVACYENG-1491. This new value will return `true` when smokescreen is actively draining connections and is meant to be used in healthcheck functions.


cds2-stripe:
This still depends on #68 getting merged, but wanted to throw it up for feedback.

r? @rlk-stripe 
cc @stripe/security-infra

cds2-stripe:
@rlk-stripe one more 👍 if you could. I removed a bunch of stale references to the old maintenance file.

PR #68, opened 2019-06-24
bump go ci version to remove 1.9 and include 1.11,1.12

cds2-stripe:
Details in tin. `go vet` was failing using 1.9 when trying to print an `atomic.Value` I added to the config struct in 1aba1bb910707c7512b3d8c84748b9bc8abb37a3. noCopy was removed from atomic.Value in go versions 1.10+.

r? @mdoan-stripe 
cc @stripe/security-infra


mdoan-stripe:
lgtm

PR #72, opened 2019-08-22
conntrack package

cds2-stripe:
This PR is the first of many working towards breaking up the monolithic Config struct. Breaking this up will make smokescreen more modular and easier to develop.

Besides addition to separating out the `conntrack` package, this PR:
* Fixes a bug where we weren't accurately counting bytes in/out
* Removes lock contention introduced by the per-connection stat tracking for every read/write
* Creates a separate `InstrumentedConnStats` reporting struct
* Removes counting `wakeups`
* Adds tests
* Changes the `IsShuttingDown` variable to `ShuttingDown`

I've been testing this most of today in QA and it works as expected.

r? @hans-stripe 
cc @stripe/security-infra


pkg/smokescreen/conntrack/instrumented_conn_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestInstrumentedConnByteCounting sends a fixed size message between two
+// instrumented connections and ensures the bytes in and out are correct.
+func TestInstrumentedConnByteCounting(t *testing.T) {
+	assert := assert.New(t)
+
+	addr := "localhost:8888"
```
hans-stripe: Could we listen on a dynamic available port here instead of hardcoding this?

pkg/smokescreen/config.go
```diff
@@ -47,12 +47,11 @@ type Config struct {
...
 	StatsSocketFileMode          os.FileMode
-	StatsServer                  *StatsServer    // StatsServer
-	ConnTracker                  *sync.Map       // The zero Map is empty and ready to use
-	IdleThresholdSec             time.Duration   // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
-	WgCxns                       *sync.WaitGroup // This wait group tracks *open* (active & idle) connections for graceful shutdown.
-	Healthcheck                  http.Handler    // User defined http.Handler for optional requests to a /healthcheck endpoint
-	IsShuttingDown               atomic.Value    // Stores a boolean value indicating whether the proxy is actively shutting down
+	StatsServer                  *StatsServer // StatsServer
+	ConnTracker                  *conntrack.Tracker
+	IdleThresholdSec             time.Duration // Consider a connection idle if it has been inactive (no bytes transferred) for this many seconds.
```
hans-stripe: No action required: I am mildly bothered by the `Sec` suffix when this is a `time.Duration` 😛

cds2-stripe: Agreed! time.Duration shouldn't be bound to seconds. Fixed.

pkg/smokescreen/conntrack/instrumented_conn_test.go
cds2-stripe: 👍

cds2-stripe:
The `wakeup` metric was a bit misleading, and instead measured the number of read/writes for a given connection. I never found much use for the metric that couldn't be inferred from the connection duration or bytes in/our.

PR #73, opened 2019-09-13
[wip] acl package

cds2-stripe:
#### Reviewers
r? 
cc 

#### Summary

Following #72 , this is the next iteration of breaking up smokescreen into component packages. This PR moves the ACL creation and decision logic into its own package.

The `acl` package:
* Creates a `Loader` interface for clients to specify from where they want their ACL configuration loaded from. Right now the only implementation is the `YAMLLoader`, but we can extend this in the future.
* Encapsulates ACL decisions into a `Result` struct
* Fixes a bug where smokescreen was incorrectly claiming to have used the default rule in an ACL decision
*  

I also discovered that the `cmd/integration_test.go` has been broken since before May. I fixed it enough to stop `go vet` from complaining but I will need to circle back and look at this.

#### Motivation

Continue to modularize the code base and chip away at paring down the monolithic `Config` struct.

#### Test plan

There were existing unit tests with good coverage. I only had to make minor modifications to the existing tests, so I'm confident that the expected functionality is correct. I also added a few additional tests for new functions.


cds2-stripe:
Closed in favor of #75

PR #75, opened 2019-09-26
smokescreen acl package

cds2-stripe:
## Reviewers
r? @charles-stripe 
cc @stripe/security-infra 

This is definitely best reviewed commit-by-commit.

## Summary
Following #72 , this is the next iteration of breaking up smokescreen into component packages. This PR moves the ACL creation and decision logic into its own package.

The `acl` package:
* Creates a `Loader` interface for clients to specify from where they want their ACL configuration loaded from. Right now the only implementation is for YAML configurations, but we can extend this in the future.
* Smokescreen uses the `Decider` interface to make ACL decisions
* The ACL package encapsulates these decisions into a `Result` struct
* Fixes a bug where smokescreen was incorrectly claiming to have used the default rule in an ACL decision

I also discovered that the `cmd/integration_test.go` has been broken since before May. I fixed it enough to stop `go vet` from complaining but I will need to circle back and look at this.

## Motivation
Continue to modularize the code base and chip away at paring down the monolithic Config struct.

## Test plan
There were existing unit tests with good coverage. I only had to make minor modifications to the existing tests, so I'm confident that the expected functionality is correct. I also added a few additional tests for new functions.

I've also brought this into `egress-proxy-srv` and had this running in QA without issue.


pkg/smokescreen/acl/v1/acl.go
```diff
@@ -0,0 +1,230 @@
...
+
+	err = acl.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	acl.Logger = logger
+
+	if acl.DefaultRule == nil {
+		acl.Warn("no default rule set")
```
charles-stripe: What happens if no default rule is set? It might be worth including that in the output here

pkg/smokescreen/acl/v1/acl.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = acl.ValidateDomains(r.DomainGlobs)
+	if err != nil {
+		return err
+	}
+
+	acl.Rules[svc] = r
```
charles-stripe: Is there ever possible that we would overwrite a pre-existing rule here? If so, is that something we should warn or error on?

pkg/smokescreen/acl/v1/acl.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// if the host matches any of the rule's allowed domains, allow
+	for _, dg := range rule.DomainGlobs {
+		if hostMatchesGlob(host, dg) {
+			d.Result, d.Reason = Allow, "host matched allowed domain in rule"
+			return d, nil
+		}
+	}
+
+	// if the host matches any of the global deny list, deny
```
charles-stripe: Just confirming: We want individual services' ACLs to override the global deny list?

pkg/smokescreen/acl/v1/acl.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return nil
+}
+
+// ValidateDomains takes a slice of domains and verifies they conform to
+// smokescreen's domain glob policy.
+//
+// Domains can only contain a single wildcard prefix
+// Domains cannot be represented as a sole wildcard
+func (acl *ACL) ValidateDomains(domains []string) error {
```
charles-stripe: Since we're iterating through a list of domains here, should we include the domains in the error messages?

pkg/smokescreen/acl/v1/policy.go
```diff
@@ -0,0 +1,43 @@
...
+type DecisionResult int
+
+const (
+	Allow DecisionResult = iota
+	AllowAndReport
+	Deny
+)
+
+func (d DecisionResult) String() string {
+	return [...]string{"Unknown", "Allow", "AllowAndReport", "Deny"}[d]
```
charles-stripe: I'm not well-versed in this notation but it looks like we may be missing an `Unknown` value for the `DecisionResult` enum

pkg/smokescreen/acl/v1/acl.go
cds2-stripe: If no default rule is set, services without a rule will be denied. ++ to updating the message!

pkg/smokescreen/acl/v1/acl.go
cds2-stripe: Nice catch. Will fix this.

pkg/smokescreen/acl/v1/acl.go
cds2-stripe: For better or worse, this is how the implementation works today https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/acl_loader_v1.go#L38. Given that this is already a rather large PR I was punting on changing the expected behavior. I do agree this is something we should revisit because I'm not sure this is the behavior we actually want.

pkg/smokescreen/acl/v1/acl.go
cds2-stripe: 👍

pkg/smokescreen/acl/v1/policy.go
cds2-stripe: You are 100% right. Fixing this actually uncovered another bug. I added some regression tests.

cds2-stripe:
PTAL? @charles-stripe

pkg/smokescreen/acl/v1/acl.go
charles-stripe: Got it, thanks for taking the time to explain it to me!

pkg/smokescreen/acl/v1/acl.go
charles-stripe: Nice fix, thanks for the additional test coverage!

PR #76, opened 2019-09-26
Remove X-Smokescreen-Trace-ID header from requests 

jjiang-stripe:
## Reviewers
r? @cds2-stripe 
cc @stripe/security-infra 

## Summary
This strips the `X-Smokescreen-Trace-ID` header from requests and adds the id as a field to the logs. 

I added the trace id as a field on the `userdata` so we didn't lose it when logging response.

## Motivation
Clients can add the `X-Smokescreen-Trace-ID` header to requests and use it to identify useful debugging information (e.g. service name) when there are egress denials. We shouldn't forward this header to the destination though so we need to remove it in the proxy and log it.

## Test plan
Unit tests!

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -272,10 +272,10 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		config.Log.WithFields(
 			logrus.Fields{
-				"source_ip":            req.RemoteAddr,
-				"requested_host":       req.Host,
-				"url":                  req.RequestURI,
-				"smokescreen_trace_id": req.Header.Get(traceHeader),
+				"source_ip":      req.RemoteAddr,
+				"requested_host": req.Host,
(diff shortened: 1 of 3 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -270,14 +272,23 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		config.Log.WithFields(
 			logrus.Fields{
-				"source_ip":      req.RemoteAddr,
-				"requested_host": req.Host,
-				"url":            req.RequestURI,
+				"source_ip":            req.RemoteAddr,
+				"requested_host":       req.Host,
+				"url":                  req.RequestURI,
+				"smokescreen_trace_id": req.Header.Get(traceHeader),
```
cds2-stripe: nit: can we change this to just `trace_id`?

jjiang-stripe:
re-r? @cds2-stripe

PR #79, opened 2019-11-07
fix sample ACL yaml paths in README

ransford-stripe:
To reflect recent yaml reorg (581f9de).


ransford-stripe:
Oops, I see #78 already addresses this!

PR #84, opened 2019-12-06
timeout idle conns

cds2-stripe:
r? @adunham-stripe 
cc @stripe/platform-security 

This PR wraps `InstrumentedConn` with an optional timeout. A deadline specified by `IdleTimeout` is added to the `net.Conn` passed to `NewInstrumentedConn`. The deadline is extended by `IdleTimeout` after every call to read/write. A zero value for the timeout duration will not associate a deadline with the connection.

We have been seeing a number of `InstrumentedConn`s persist for hours past their last read/write time. This helps protect us from misbehaving clients/servers which do not correctly close their connection.

### DIFF
--- a/pkg/smokescreen/conntrack/instrumented_conn.go
+++ b/pkg/smokescreen/conntrack/instrumented_conn.go
@@ -48,6 +48,9 @@ func (t *Tracker) NewInstrumentedConn(conn net.Conn, traceId, role, outboundHost
 		BytesOut:     &bytesOut,
 	}
 
+	if t.IdleTimeout != 0 {
+		ic.Conn.SetDeadline(time.Now().Add(t.IdleTimeout))
+	}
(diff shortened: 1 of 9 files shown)


cds2-stripe:
re-r? @adunham-stripe

pkg/smokescreen/conntrack/instrumented_conn.go
```diff
@@ -48,6 +48,9 @@ func (t *Tracker) NewInstrumentedConn(conn net.Conn, traceId, role, outboundHost
 		BytesOut:     &bytesOut,
 	}
 
+	if t.IdleTimeout != 0 {
+		ic.Conn.SetDeadline(time.Now().Add(t.IdleTimeout))
```
hans-stripe: Nit: Can you reuse the `time.Now()` call above?

PR #86, opened 2019-12-18
Run gofmt with simplify

ferhatelmas:


mattm-stripe:
CLA not signed; I'm going to close this as stale.
However I am working on improving CI so that we can enforce formatting and so the concept here won't be lost

PR #94, opened 2020-01-22
golint clean

ransford-stripe:
So as not to annoy upstream dependents.


ransford-stripe:
Oops, missed a few (classic `golint cmd pkg` vs. `golint cmd/... pkg/...` error); don't merge yet.

PR #99, opened 2020-02-14
default to exit_timeout for shutdown timeout

cds2-stripe:
If `idle_timeout` is set to 0, an exit signal to Smokescreen will immediately kill all connections and bypass the configured `exit_timeout`. This PR ensures that if `idle_timeout` is 0 Smokescreen waits at least `exit_timeout` before terminating any active connections.

I also added additional logging so that it's easier to determine the state of the connections when Smokescreen terminates.

r? @stripe/platform-security


cds2-stripe:
Trying to raise this from the dead as I'm pushing on some other egress work!

r? @adunham-stripe

PR #100, opened 2020-04-01
improve smokescreen error handling

cds2-stripe:
This PR fixes and improves the ways Smokescreen returns errors to proxy clients.

For traditional HTTP requests, we were incorrectly deleting the `X-Smokescreen-Error` in cases where Smokescreen's `net.Transport` timed out dialing the remote host.  This was fixed by checking for an error which is attached to the proxy's context.

For CONNECT requests, we weren't setting the  `X-Smokescreen-Error` header for anything but ACL denials. To fix this required hooking into our forked version of `goproxy` to invoke an HTTP error handler to create a custom response generated using a call to `rejectResponse`. Previously, goproxy would have simply 502d the request without an associated error header. This was fixed here https://github.com/stripe/goproxy/pull/11.

Finally, this PR improves the errors returned by Smokescreen. Previously, Smokescreen would return a `407 Proxy Authentication Required` for _everything_ under its control. In cases where the connection timed out or failed to establish, goproxy would return a 502.

We now return the following status codes:
* 407 Proxy Authentication Required - for requests which are denied by the ACL.
* 500 Internal Server Error - a catch all for errors internal to Smokescreen.
* 502 Bad Gateway - anything which returns a `net.Error` that is not related to a timeout (DNS resolution failures, etc).
* 504 Gateway Timeout - returned when the connection times out when being established. This timeout duration is configured by the `ConnectTimeout` field and used by `net.DialTimeout`.

I don't plan on rolling this out until we land the blocked timeout changes, but wanted to get this up for PR as it's impacting some of our internal users.

r? @rwg-stripe 
cc @stripe/platform-security 
cc @tmaxwell-stripe


pkg/smokescreen/smokescreen.go
```diff
@@ -308,9 +339,17 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 	// Handle traditional HTTP proxy
 	proxy.OnRequest().DoFunc(func(req *http.Request, pctx *goproxy.ProxyCtx) (*http.Request, *http.Response) {
 		// Attach smokescreenContext to goproxy.ProxyCtx
+		// We are intentionally *not* setting pctx.HTTPErrorHandler because with traditional HTTP
+		// proxy requests we are able to speficy the request during the call to OnResponse().
```
rwg-stripe, line 343: typo: `speficy`

cds2-stripe:
ptal? @rwg-stripe

cds2-stripe:
re-r? @rwg-stripe 

I think this is ready to go. I added a terminal log line to log error messages, and removed the log event in the HTTP `OnResponse` handler which may result in errors being double logged.

pkg/smokescreen/smokescreen.go
```diff
@@ -258,29 +259,61 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+			code = http.StatusGatewayTimeout
+			msg = "Timed out connecting to remote host: " + e.Error()
+		} else {
+			status = "Bad gateway"
+			code = http.StatusBadGateway
+			msg = "Failed to connect to remote host: " + e.Error()
+		}
+	} else if e, ok := err.(denyError); ok {
+		status = "Request rejected by proxy"
+		code = http.StatusProxyAuthRequired
```
ransford-stripe, line 296: I'm not convinced 407 Proxy Authentication Required is better than 403 Forbidden. A proxy server sending 407 is supposed to also send a `Proxy-Authenticate` header with a suggestion for the client (like "try Basic auth"), and the client is supposed to be able to interpret that suggestion and retry its request if it wants to. 403 is more like "nope, don't try again."

If you want to stick with 407, I suggest including a comment justifying that choice.

pkg/smokescreen/smokescreen.go
```diff
@@ -258,29 +259,61 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+		msg = fmt.Sprintf(denyMsgTmpl, pctx.Req.Host, e.Error())
+	} else {
+		status = "Internal server error"
+		code = http.StatusInternalServerError
+		msg = "An unexpected error occurred: " + err.Error()
+		sctx.cfg.Log.WithFields(logrus.Fields{
```
ransford-stripe, line 302: `sctx.logger.WithFields(...)`?

pkg/smokescreen/smokescreen.go
```diff
@@ -386,26 +427,45 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
...
 		err := handleConnect(config, pctx)
 		if err != nil {
-			pctx.Resp = rejectResponse(pctx.Req, config, err)
+			pctx.Resp = rejectResponse(pctx, err)
 			return goproxy.RejectConnect, ""
 		}
 		return goproxy.OkConnect, host
 	})
 
+	// Strangely, goproxy can invoke this same function twice for a single HTTP request.
```
ransford-stripe, line 445: 😬

ransford-stripe: This handler is _very_ weird and hard to reason about.

pkg/smokescreen/smokescreen_test.go
```diff
@@ -106,38 +106,65 @@ func TestClassifyAddr(t *testing.T) {
 func TestClearsErrorHeader(t *testing.T) {
```
ransford-stripe, line 106: Tests pathological behavior -> merits a comment explaining what it's testing.

pkg/smokescreen/smokescreen_test.go
```diff
@@ -528,12 +556,66 @@ func TestProxyTimeouts(t *testing.T) {
...
+		proxy := proxyServer(cfg)
+		remote := httptest.NewServer(h)
+		client, err := proxyClient(proxy.URL)
+		r.NoError(err)
+
+		req, err := http.NewRequest("GET", remote.URL, nil)
+		r.NoError(err)
+
+		resp, _ := client.Do(req)
+		r.Equal(http.StatusGatewayTimeout, resp.StatusCode)
```
ransford-stripe, line 616: Nice!

pkg/smokescreen/smokescreen.go
cds2-stripe: It is! And every time I need to come back to working on something involving this flow I spend too much time reminding myself how the goproxy machinery works. Hopefully this will help the next fortunate soul stumble through this code.

pkg/smokescreen/smokescreen.go
cds2-stripe: `407 Proxy Authentication Required` has been the return code for Smokescreen ACL denials since its inception. I completely agree about the 403 being a more appropriate status code, but I guarantee there are now services relying on the old status code. I'd prefer to save this for another migration. I'll also open an issue to track it.

pkg/smokescreen/smokescreen.go
cds2-stripe: 👍

pkg/smokescreen/smokescreen_test.go
cds2-stripe: Done!

pkg/smokescreen/testdata/acl.yaml
```diff
@@ -5,7 +5,7 @@
         project: security
         action: enforce
         allowed_domains:
-          - notarealhost.com
+          - notarealhost.notarealtld
```
rwg-stripe, line 8: While it's unlikely that `.notarealtld` will ever become a real gTLD, `.test` exists explicitly for testing and will never have any hostnames/domains in it.

PR #101, opened 2020-04-08
migrate to go modules

### LINKED ISSUE #82: Add Go modules support

cds2-stripe:
r? @adunham-stripe 
cc @stripe/platform-security 

This PR migrates Smokescreen away from `dep` to Go modules for dependency management. Smokescreen will also now only support the two most recent releases of Go. Currently this is `1.13` and `1.14`. The Travis build configuration was updated to test with all of the minor versions of these releases.

I also changed Smokescreen to import our forked version of `goproxy` directly. Previously we were rewriting the import path to point to our forked version.

I was able to successfully build our egress proxies off of this branch.

Closes #82.


cds2-stripe:
@adunham-stripe could you take a look? I've updated `stripe/goproxy` to remove import references to the upstream with https://github.com/stripe/goproxy/pull/14.

PR #102, opened 2020-04-16
fix integration tests

cds2-stripe:
This PR fixes our long forgotten integration tests. They've been broken since I started working on Smokescreen and were never included as part of the CI pipeline.

There were a number of latent bugs and race conditions that made fixing these tests quite difficult.
* By not reading the response from the initial CONNECT request we were actually creating a race condition where the response was returned unsolicited to the returned HTTP client. These surfaced as runtime errors: `Unsolicited response received on idle HTTP channel`.
* It turns out setting and unsetting shared environment variables during tests is also not a good idea. Some of these tests depend on state in `http_proxy` and `https_proxy` and would race with one another. I've side stepped this by breaking out the dependent tests into their own test commands.
* We were incorrectly assuming that the `index-1` last log entry was the entry for `CANONICAL-PROXY-DECISION`. This was another race against the `CANONICAL-PROXY-CN-CLOSE` log line.

I also added comments to describe the intentions behind the checks and setup routines to hopefully make this easier for the next person who has to work with these tests. They still aren't perfect but should be in a much better state.

r? @ransford-stripe 
cc @stripe/platform-security

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -132,12 +132,12 @@ func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error,
 	}
 
 	if len(entries) > 0 {
-		lastEntryIndex := len(entries) - 1
-		entry := entries[lastEntryIndex]
+		entry := findLogEntry(entries, smokescreen.LOGLINE_CANONICAL_PROXY_DECISION)
+		a.NotNil(entry)
 		a.Equal(entry.Message, smokescreen.LOGLINE_CANONICAL_PROXY_DECISION)
 
 		a.Contains(entry.Data, "allow")
-		a.Equal(test.ExpectAllow, entries[lastEntryIndex].Data["allow"])
+		a.Equal(test.ExpectAllow, entry.Data["allow"])
 
 		a.Contains(entry.Data, "proxy_type")
 		if test.OverConnect {
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -289,7 +289,7 @@ func TestSmokescreenIntegration(t *testing.T) {
 		proxyServers[useTLS] = proxyServer
 
 		if useTLS {
-			externalHosts[useTLS] = "https://api.github.com:443"
+			externalHosts[useTLS] = "https://api.stripe.com:443"
 
 			httpServer := httptest.NewTLSServer(ProxyTargetHandler)
 			defer httpServer.Close()
@@ -373,6 +373,10 @@ func TestSmokescreenIntegration(t *testing.T) {
 
 					if expectAllow {
 						testCase.ExpectStatus = http.StatusOK
+						if overTLS && !authorizedHost {
+							// The Stripe API returns a 403 to a bare HTTP GET request
+							testCase.ExpectStatus = http.StatusUnauthorized
+						}
 					} else {
 						testCase.ExpectStatus = http.StatusProxyAuthRequired
 					}
@@ -470,10 +474,11 @@ func TestInvalidUpstreamProxyConfiguration(t *testing.T) {
 			var proxyTarget string
 			var upstreamProxy string
 
-			// These proxy targets don't actually matter, as
+			// These proxy targets don't actually matter as the requests wont be sent.
+			// because the resolution of the upstream proxy will fail.
 			if overConnect {
 				upstreamProxy = "https://notaproxy.prxy.svc:443"
-				proxyTarget = "https://api.github.com:443"
+				proxyTarget = "https://api.stripe.com:443"
 			} else {
 				upstreamProxy = "http://notaproxy.prxy.svc:80"
 				proxyTarget = "http://checkip.amazonaws.com:80"
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -3,10 +3,11 @@
 package cmd
 
 import (
-	"bytes"
+	"bufio"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -18,6 +19,7 @@ import (
 	"os"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/go-cleanhttp"
 	"github.com/sirupsen/logrus"
@@ -32,7 +34,8 @@ import (
 type DummyHandler struct{}
 
 func (s *DummyHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	io.WriteString(rw, "ok")
+	time.Sleep(50 * time.Millisecond)
+	io.WriteString(rw, "okok")
 }
 
 func NewDummyServer() *http.Server {
@@ -58,12 +61,15 @@ func testRFRCert(req *http.Request) (string, error) {
 
 type TestCase struct {
 	ExpectAllow   bool
+	Action        acl.EnforcementPolicy
+	ExpectStatus  int
 	OverTls       bool
 	OverConnect   bool
 	ProxyURL      string
 	TargetPort    int
 	RandomTrace   int
-	Host          string
+	TargetHost    string
+	TargetScheme  string
 	RoleName      string
 	UpstreamProxy string
 }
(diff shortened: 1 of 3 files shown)


cmd/integration_test.go
```diff
@@ -306,85 +328,109 @@ func TestSmokescreenIntegration(t *testing.T) {
 
 	var testCases []*TestCase
 
+	// This generates all the permutations for the common test cases
```
ransford-stripe, line 331: A little more commentary here would be helpful, especially since inside the nested loop we decide that some of the test cases aren't interesting/meaningful.

cmd/integration_test.go
```diff
@@ -306,85 +328,109 @@ func TestSmokescreenIntegration(t *testing.T) {
...
 		// this test will fail if we're not respecting the UpstreamProxy setting
 		// and instead going straight to this host.
-		proxyCase.Host = "aws.s3.amazonaws.com"
-		proxyCase.UpstreamProxy = outsideListenerUrl.String()
-		proxyCase.ExpectAllow = true
-		proxyCase.RoleName = generateRoleForAction(acl.Open)
+		if overConnect {
+			proxyCase := baseCase
+			proxyCase.TargetScheme = "https"
+			proxyCase.TargetHost = "aws.s3.amazonaws.com"
```
ransford-stripe, line 423: What if https://aws.s3.amazonaws.com/ starts returning 200 or 301->200 or something? IMO it would be better to stand up another temporary HTTP server (as part of test setup) that returns 403 or similar.

cmd/integration_test.go
```diff
@@ -395,19 +441,42 @@ func TestSmokescreenIntegration(t *testing.T) {
 			conformResult(t, testCase, resp, err, logHook.AllEntries())
 		})
 	}
+}
+
+// This test must be run with a separate test command as the environment variables
+// required can race with the test above.
+func TestUpstreamProxySmokescreenIntegration(t *testing.T) {
```
ransford-stripe, line 448: This just tests that ill-formed proxy configurations elicit meaningful error messages, right? If so, `TestInvalidProxyConfiguration` or similar would be more descriptive.

cmd/integration_test.go
```diff
@@ -306,85 +328,109 @@ func TestSmokescreenIntegration(t *testing.T) {
...
+		}
 
 		testCases = append(testCases,
-			&unknownRoleAllowCase, &unknownRoleDenyCase,
-			&noRoleAllowCase, &noRoleDenyCase,
+			&unknownRoleDenyCase, &noRoleDenyCase,
 			&badIPRangeCase, &badIPAddressCase,
-			&proxyCase,
 		)
 	}
```
ransford-stripe, line 436: For the test cases below here, it would be good to use a meaningful name instead of `""` as the first argument to `t.Run()`. How about adding a `TestCase.String()` method that encodes test case parameters in a readable way?

cmd/integration_test.go
ransford-stripe: Also the block of `*Domain` variables above merits a comment for readability.

cmd/integration_test.go
cds2-stripe: I removed this test case, but I'm generally in favor of keeping reasonably stable external domains as this is an integration test suite which should be tested against non-local HTTP servers.

cmd/integration_test.go
cds2-stripe: Prior to every test running we print out the test case parameters. Because we can't generate meaningful `TestCase.String()` output for the autogenerated tests I'm inclined to leave this as is for now.

cmd/integration_test.go
```diff
@@ -329,25 +322,33 @@ func TestSmokescreenIntegration(t *testing.T) {
...
-			// Do not support this test case
-			if overTls && !overConnect {
+	// * proxy requests using CONNECT and regular HTTP proxy
+	// * TLS and non-TLS proxy targets
+	// * hosts authorized and not authorized by the config in testdata/sample_config.yaml
+	// * enforce, report, and open enforcement policies
+	for _, overConnect := range overConnectOptions {
+		for _, overTLS := range overTLSOptions {
+			// Explicitly do not support these test cases
+			// 1) You cannot tunnel TLS using a traditional HTTP proxy request
```
ransford-stripe, line 332: This is really clear now!

cmd/integration_test.go
```diff
@@ -356,77 +357,68 @@ func TestSmokescreenIntegration(t *testing.T) {
...
-						OverConnect:  overConnect,
-						ProxyURL:     servers[overTls].URL,
-						TargetPort:   port,
-						TargetHost:   host,
-						TargetScheme: targetScheme,
-						RoleName:     generateRoleForAction(action),
+					if expectAllow {
+						testCase.ExpectStatus = http.StatusOK
+						if overTLS && !authorizedHost {
+							// The Stripe API returns a 403 to a bare HTTP GET request
```
ransford-stripe, line 377: Nice, this is a great compromise — we _do_ control this condition, or at least we can poke someone who does!

cmd/integration_test.go
```diff
@@ -356,77 +357,68 @@ func TestSmokescreenIntegration(t *testing.T) {
...
+		unknownRoleDenyCase.ExpectStatus = http.StatusProxyAuthRequired
 
 		// This must be a global unicast, non-loopback address or other IP rules will
 		// block it regardless of the specific configuration we're trying to test.
 		badIPRangeCase := baseCase
-		badIPRangeCase.TargetHost = "1.1.1.1"
+		badIPRangeCase.TargetURL = "http://1.1.1.1:80"
 		badIPRangeCase.ExpectAllow = false
-		badIPRangeCase.RoleName = generateRoleForAction(acl.Open)
+		badIPRangeCase.ExpectStatus = http.StatusProxyAuthRequired
```
ransford-stripe, line 412: This is good. The test makes a precise expectation clear.

cmd/integration_test.go
```diff
@@ -438,48 +430,77 @@ func TestSmokescreenIntegration(t *testing.T) {
...
 	// handled error so it would fail many of the checks in our other tests. We really
 	// only care to ensure that these requests never succeed.
 	for _, overConnect := range []bool{true, false} {
 		t.Run(fmt.Sprintf("illegal proxy with CONNECT %t", overConnect), func(t *testing.T) {
-			targetScheme := "http"
-			targetPort := 80
+			var proxyTarget string
+			var upstreamProxy string
+
+			// These proxy targets don't actually matter as the requests wont be sent.
```
ransford-stripe, line 477:
```suggestion
			// These proxy targets don't actually matter as the requests won't be sent.
```

cmd/integration_test.go
```diff
@@ -63,31 +54,17 @@ type TestCase struct {
...
-	entry := findLogEntry(logs, "CANONICAL-PROXY-DECISION")
-	r.NotNil(entry)
-	// Make sure the hostname resolution of the upstream failed. This indicates
-	// that the request wasn't forwarded directly to the proxy target.
-	a.Contains(entry.Data["error"], "notaproxy.service")
-}
-
-func conformResult(t *testing.T, test *TestCase, resp *http.Response, err error, logs []*logrus.Entry) {
+// isValidProxyResponse validates tests cases and expected responses from TestSmokescreenIntegration
+func isValidProxyResponse(t *testing.T, test *TestCase, resp *http.Response, err error, logs []*logrus.Entry) {
```
ransford-stripe, line 67: Ah, I thought you wanted to use this function in a condition, like `if isValidProxyResponse { ... }`, but if not, I retract my suggestion to call it `is*` (which looks like it should return a boolean).

cmd/integration_test.go
```diff
@@ -438,48 +430,77 @@ func TestSmokescreenIntegration(t *testing.T) {
...
 		})
 	}
 }
 
+// isValidProxyResponseWithUpstream validates tests cases and expected responses
+// from TestUpstreamProxySmokescreenIntegration. This validates that requests
+// sent to a smokescreen instance with an additional upstream proxy set
+// (proxy chaining) forwards the request to the next proxy hop instead of directly
+// to the proxy target.
+func isValidProxyResponseWithUpstream(t *testing.T, test *TestCase, resp *http.Response, err error, logs []*logrus.Entry) {
```
ransford-stripe, line 443: Ditto my earlier suggestion — drop the `is*` prefix if you're not returning a boolean.

cmd/integration_test.go
cds2-stripe: Agreed! I changed these to `validateProxyResponse...`

cmd/integration_test.go
cds2-stripe: Changed to `validateProxyResponseWithUpstream`

cds2-stripe:
ptal? @ransford-stripe

PR #106, opened 2020-05-06
use github actions for ci

cds2-stripe:
r? @stripe/platform-security 

This deprecates TravisCI in favor of Github Actions. The `go test` parameters are unchanged.


.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
+on: [push, pull_request]
```
rwg-stripe, line 1: I'm surprised this works — `on` is a "truthy" value in YAML and should get treated like `true` when it's not in quotes...

```
Python 3.7.7 (default, Mar 10 2020, 15:43:33) 
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import yaml, sys
>>> yaml.dump(yaml.load("on: [push, pull_request]\n"), stream=sys.stdout)
true: [push, pull_request]
>>> yaml.dump(yaml.load("'on': [push, pull_request]\n"), stream=sys.stdout)
'on': [push, pull_request]
```

However, all of GitHub's documentation shows `on` being used unquoted in workflow YAML files, so 🤷‍♂️...

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+        platform: [ubuntu-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Setup env
+      run:
+        echo ::set-env name=GOFLAGS::-mod=vendor
```
rwg-stripe, line 17: Would you be willing to change this (and the same thing under `name: Vet` below) to look either like...

```
      run: 'echo ::set-env name=GOFLAGS::-mod=vendor'
```

...or...

```
      run: |
        echo ::set-env name=GOFLAGS::-mod=vendor
```

...?  I'm afraid if someone needs to add a second command to run, they'll add it on a new line under the existing command, and then the two lines will just get squashed together into one line instead of being treated as separate lines.

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
+on: [push, pull_request]
+name: Test
+jobs:
+  test:
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x]
+        platform: [ubuntu-latest]
```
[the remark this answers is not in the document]
cds2-stripe: This is a great point! It's basically free to test against `ubuntu-16.04` so I've added it.

.github/workflows/test.yml
cds2-stripe: 👍

cds2-stripe:
Thanks for the great feedback!

ptal? @rwg-stripe

PR #107, opened 2020-05-11
small collection of logging improvements

cds2-stripe:
This PR contains a number of small improvements to Smokescreen's logging:
* Use a request scoped `logrus.Entry` attached to the request's context to log common fields. This logger is shared with `InstrumentedConn`s to ensure that field names are consistent.
* Added an `id` field to simplify pairing `CANONICAL-PROXY-DECISION` and `CANONICAL-PROXY-CN-CLOSE` events.
* Lifted `CONNECT` logging calls to the `OnConnect` handler to match how HTTP proxy requests are logged.
* Removed `trace_id` as a struct member, as it is now passed by the `logrus.Entry`.
* Removed the `idle` field logged during `CANONICAL-PROXY-CN-CLOSE` events as this field was unintuitive and not actually indicative of the connection being idle. It was replaced by the field `active_at_termination` which will only be emitted if Smokescreen is in the process of shutting down.
* Added a `last_activity` timestamp which is logged for every `CANONICAL-PROXY-CN-CLOSE`.

There were existing tests which hooked into logrus to check that the correct log fields are being emitted.

r? @hans-stripe 
cc @stripe/platform-security

### DIFF
--- a/pkg/smokescreen/conntrack/instrumented_conn.go
+++ b/pkg/smokescreen/conntrack/instrumented_conn.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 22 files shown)


pkg/smokescreen/conntrack/instrumented_conn.go
```diff
@@ -112,18 +111,16 @@ func (ic *InstrumentedConn) Close() error {
...
-		"start_time":  ic.Start.UTC(),
-		"end_time":    end.UTC(),
-		"duration":    duration,
-		"timed_out":   timeout,
-		"error":       errorMessage,
+	ic.logger.WithFields(logrus.Fields{
+		"bytes_in":      ic.BytesIn,
+		"bytes_out":     ic.BytesOut,
+		"role":          ic.Role,
+		"remote_addr":   ic.Conn.RemoteAddr(),
```
hans-stripe, line 118: Can this be changed to match the `dst_ip` / `dst_port` pair in the decision log?

PR #109, opened 2020-05-13
add logging for failed net.Dials

cds2-stripe:
While working on runbooks for some of our detectors I realized we didn't have good logs which are generated when we fail to establish a connection to the proxy target. We increment the `smokescreen.cn.atpt.fail.total` metric, and the go runtime will emit and error, but that's about it. For traditional HTTP proxy requests, we also catch and log this as a generic error in the `OnResponse()` function.

Even with the error handling improvements queued up in https://github.com/stripe/smokescreen/pull/100/, we are only returning the failure to the client and not logging any useful debugging information.

This makes it possible to tie `net.Dial` failures to individual requests.

r? @rwg-stripe 
cc @stripe/platform-security


cds2-stripe:
Added this functionality to #100 .

PR #111, opened 2020-05-29
goproxy: close after copy

cds2-stripe:
This brings in https://github.com/stripe/goproxy/pull/15 from `stripe/goproxy`. The PR reverts the way we handle connection closures to how it was being handled previously.

Additionally, this PR:
* Adds a regression tests to ensure Smokescreen doesn't leak connections when a TCP connection is half-closed.
* Fixes a data race found by Go's race detector.
* Removes the `timed_out` field on `CANONICAL-PROXY-CN-CLOSE` log events. See the comment on the goproxy PR, but we're no longer able to rely on this field being accurate due to a race with the runtime. I believe it's better to not surface this field if it cannot reliably report on whether a connection was timed out. It should still be possible to infer if a connection was timed out by comparing the `last_activity` field to the configured inactivity timeout.

r? @hans-stripe 
cc @stripe/platform-security

### DIFF
--- a/pkg/smokescreen/smokescreen_test.go
+++ b/pkg/smokescreen/smokescreen_test.go
@@ -5,6 +5,7 @@ package smokescreen
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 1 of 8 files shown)


pkg/smokescreen/smokescreen_test.go
```diff
@@ -625,6 +626,64 @@ func TestProxyTimeouts(t *testing.T) {
 	})
 }
 
+// TestProxyHalfClosed tests that the proxy and proxy client correctly
+// closes all connections if the proxy target attempts to half-close
+// the TCP connection.
+func TestProxyHalfClosed(t *testing.T) {
```
hans-stripe, line 632: I like this test! What do you think about adding a second test for a proxy _client_ closing its write socket? That would cover both cases and make sure we don't end up with regressions.

PR #112, opened 2020-06-05
Log amount of time net.DialTimeout takes to establish the connection

anktjsh-stripe:
This logs the host url with the amount of time the DialTimeout function takes to establish a connection to that url. 

r? @cds2-stripe  
cc @stripe/platform-security

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -48,11 +48,12 @@ type aclDecision struct {
 }
 
 type smokescreenContext struct {
-	cfg       *Config
-	start     time.Time
-	decision  *aclDecision
-	proxyType string
-	logger    *logrus.Entry
+	cfg           *Config
+	start         time.Time
+	decision      *aclDecision
+	proxyType     string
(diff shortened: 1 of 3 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -53,6 +53,7 @@ type smokescreenContext struct {
 	decision  *aclDecision
 	proxyType string
 	logger    *logrus.Entry
+	requestedHost string
```
cds2-stripe, line 56: This needs to be `gofmt`d, and we should be failing commits in CI which haven't. If you're interested in adding this to the Github Actions please do!

pkg/smokescreen/smokescreen.go
```diff
@@ -349,6 +356,7 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
 		logger:    logger,
 		proxyType: proxyType,
 		start:     start,
+		requestedHost: req.Host,
```
cds2-stripe, line 359: Should also `gofmt` this!

PR #116, opened 2020-06-26
Add flag for proxy protocol if using smokescreen as a reverse proxy

anktjsh-stripe:
Support proxy protocol for reverse proxy

#### Reviewers
r? @hans-stripe 
cc @stripe/platform-security  

#### Summary
Enable the proxy protocol based on a config flag when smokescreen is being used as a reverse proxy

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -255,7 +255,11 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 		return nil, err
 	}
 	sctx.cfg.StatsdClient.Incr("cn.atpt.success.total", []string{}, 1)
-
+	
+	if sctx.cfg.ReverseProxyProtocol {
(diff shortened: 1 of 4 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -255,7 +255,11 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 		return nil, err
 	}
 	sctx.cfg.StatsdClient.Incr("cn.atpt.success.total", []string{}, 1)
-
+	
+	if sctx.cfg.ReverseProxyProtocol {
+		conn = proxyproto.NewConn(conn, sctx.cfg.ConnectTimeout)
```
hans-stripe, line 260: Looking at the library, I don't think this will do what we want. `NewConn` creates a new connection that listens for an incoming proxy protocol header (i.e. from the host we're connecting to), rather than sending a proxy protocol header onwards.

PR #118, opened 2020-08-31
Update README with current help contents

alexmv:


hans-stripe (approved):
Thanks for helping keep this up to date!

PR #120, opened 2020-08-31
Update comment about how graceful socket closing works

alexmv:
This comment was originally written in e1ec5a5b3cf2 and was not
updated when the graceful shutdown code as overhauled in 804861da2e37.

----

I'm not actually confident in the rationale that I updated this to.  I'd love to hear more about the reason, if what I wrote is not accurate.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -640,10 +640,11 @@ func runServer(config *Config, server *http.Server, listener net.Listener, quit 
 	// 1. close the listening socket (so that we don't accept *new* connections)
 	// 2. close *existing* keepalive connections once they become idle
 	//
-	// It is impossible to close existing keepalive connections, because goproxy
-	// hijacks the socket and doesn't tell us when they become idle. So all we
-	// can do is close the listening socket when we receive a signal, not accept
-	// new connections, and then exit the program after a timeout.
+	// goproxy hijacks the socket and interferes with goji's ability to do the
+	// latter.  We instead pass InstrumentedConn objects, which wrap net.Conn,
+	// into goproxy.  ConnTracker keeps a reference to these, which allows us to
+	// know exactly how long to wait until the connection has become idle, and
+	// then Close it.
 
 	if len(config.StatsSocketDir) > 0 {
 		config.StatsServer = StartStatsServer(config)


cds2-stripe:
Hi there! Your rationale is correct. We pass wrapped `net.Conn`s to `goproxy` for a couple of reasons, including retaining a reference so that we can forcefully call `Close()`. I think it may be worth clarifying that we pass these wrapped connections, which are maintained by `ConnTracker`, to `goproxy` and that this is what allows us to support graceful termination.

Thank you for contributing to Smokescreen!

alexmv:
Thanks for the context!  Updated the wording.

I'm a little curious what the goji <-> goproxy bad interaction that's alluded to is -- that's the part of the comment that feels opaque to me.

alexmv:
Ping on this?

pkg/smokescreen/smokescreen.go
```diff
@@ -640,10 +640,11 @@ func runServer(config *Config, server *http.Server, listener net.Listener, quit
...
 	// 2. close *existing* keepalive connections once they become idle
 	//
-	// It is impossible to close existing keepalive connections, because goproxy
-	// hijacks the socket and doesn't tell us when they become idle. So all we
-	// can do is close the listening socket when we receive a signal, not accept
-	// new connections, and then exit the program after a timeout.
+	// goproxy hijacks the socket and interferes with goji's ability to do the
+	// latter.  We instead pass InstrumentedConn objects, which wrap net.Conn,
+	// into goproxy.  ConnTracker keeps a reference to these, which allows us to
+	// know exaclty how long to wait until the connection has become idle, and
```
cds2-stripe, line 646: exaclty -> exactly

cds2-stripe (changes requested):
Sorry for the delay, I missed the last update.

alexmv:
Updated; thanks for that catch?

Did you have thoughts on this:
> I'm a little curious what the goji <-> goproxy bad interaction that's alluded to is -- that's the part of the comment that feels opaque to me.

cds2-stripe:
@alexmv I was able to dig up a comment from Carl, who is also the author of the goji libraries, explaining why the interaction doesn't work as expected: https://github.com/stripe/smokescreen/pull/5#issuecomment-272520503

Hope this helps!

cds2-stripe (approved):

PR #122, opened 2020-09-02
goproxy: don't log use of closed network connection errors

cds2-stripe:
This bumps our version of `stripe/goproxy` to `ac962cdd7165858b68aed0c0b761f7afcf22f527` to bring in https://github.com/stripe/goproxy/pull/17.

The `stripe/goproxy` PR stops logging "use of closed network connection" errors.

r? @hans-stripe 
cc @stripe/platform-security


cds2-stripe:
OK! I fixed the CONNECT timeout test which was relying on the nondeterministic behavior we removed with the Goproxy PR.

What was happening was that the test was relying on there **always** being some kind of error returned when there was a connection timeout. Unfortunately, as is described by https://github.com/stripe/goproxy/blob/master/https.go#L150 we cannot guarantee which error that will be passed back to Smokescreen. Now that we are ignoring "used of closed network connection errors", we will nondeterministically be passed back the timeout error which is causing the flakiness in the tests.

There's a way to fix this which I described in a comment about the test, but the fix is outside the scope of this PR and I'm comfortable with the `EOF` check on the client response.

r? @hans-stripe

hans-stripe (approved):

PR #123, opened 2020-09-03
Fix up Smokescreen build version

hans-stripe:
Inspired by #121, this PR removes the not-working-as-intended `$Id$` from the version number and instead replaces it with a variable that can be easily set in the Golang build toolchain. We're also exporting the variable so it can be set from outside of Smokescreen (i.e. in our internal wrapper which imports this package).

I've also bumped the version to `0.0.2` since it's been a while.

r? @cds2-stripe 
cc @stripe/security-infra @alexmv

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,0 @@
-pkg/smokescreen/constants.go ident
--- a/pkg/smokescreen/constants.go
+++ b/pkg/smokescreen/constants.go
@@ -5,7 +5,7 @@ import (
 	"regexp"
 )
 
-const versionSemantic = "0.0.1"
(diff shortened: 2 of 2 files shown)


cds2-stripe (approved):

alexmv:
👍 

I think it might be good if the `go build -ldflags='-X github.com/stripe/smokescreen/pkg/smokescreen.VersionID=33955a3' ` example got moved out of just a comment, and into the README where it might be more discoverable.

Maybe in the form:
```
go build -ldflags="-X github.com/stripe/smokescreen/pkg/smokescreen.VersionID=$(git rev-parse --short=12 HEAD)" .
```

PR #124, opened 2020-11-10
Update README badge to use github actions test workflow

jjiang-stripe:
r? @cds2-stripe 
cc @stripe/platform-security 

The current badge is still pointing to travis, which we no longer use. I grabbed this badge off https://github.com/stripe/smokescreen/actions?query=workflow%3ATest (three dots next to search bar -> Create status badge) and set it to master pushes, which should now show us updated test status!


jjiang-stripe:
My prettier was not happy with our formatting. I guess these whitespace changes are an improvement?

cds2-stripe (approved):
Nice! Thanks for finding this.

PR #125, opened 2020-11-10
Add code coverage to our test workflow w/ goveralls

jjiang-stripe:
r? @cds2-stripe 
cc @stripe/platform-security 

This PR adds `goveralls` to our test workflow so we can generate code coverage reports. `goveralls` is the Go integration for coveralls.io, which I think just converts Go's coverage output to the input that coveralls likes. We already use coveralls for [several Stripe repos](https://coveralls.io/github/stripe/). I copied the config from https://github.com/mattn/goveralls#github-actions and the build appears at https://coveralls.io/github/stripe/smokescreen.

I also noticed some deprecation warnings for our usage of set-env (https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) and fixed those according to https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

I'm not sure if this will trigger PR comments since I'm not directly using coveralls' github action, but let's see!

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
         go-version: ${{ matrix.go-version }}
     - name: Setup env
       run: |
-        echo ::set-env name=GOFLAGS::-mod=vendor
+        echo "{GOFLAGS}={-mod=vendor}" >> $GITHUB_ENV
     - name: Checkout code
       uses: actions/checkout@v2
     - name: Vet
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,8 +23,16 @@ jobs:
     - name: Test
       run: |
         go mod verify
-        go test -race -v -timeout 2m -failfast ./... -tags=nointegration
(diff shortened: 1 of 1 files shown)


jjiang-stripe:
I don't know what type of spell I need to cast to get coveralls to comment on our PRs. 😢 


the goveralls repo has it configured on their PRs, but I don't see anything special in their [YAML file](https://github.com/mattn/goveralls/blob/master/.github/workflows/main.yml), but maybe I'm missing something?

jjiang-stripe:
coveralls only leaves comment if it had a build to compare to (makes sense!) 
I made a test PR off this branch and now coverage bot and I are on speaking terms ❤️  see #127

cds2-stripe:
This is great! I looked through the capabilities provided by `secrets.GITHUB_TOKEN` and there wasn't anything particularly concerning. This is also the suggested way to provide repository-scoped access to Actions workflows.

cds2-stripe (approved):

PR #129, opened 2020-11-27
Update tested Go versions to 1.14 and 1.15

mattm-stripe:
We should be testing on the two most recent Go versions


cds2-stripe (approved):
🎉 thank you for doing this!

PR #130, opened 2020-12-01
Make smokescreen clean of staticcheck and gosimple lints

mattm-stripe:
This cleans up a bunch of issues as found by the staticcheck and gosimple linters, called via golangci-lint which I'd like to eventually incorporate as a CI check.

For the most part there shouldn't be any functional changes here.  Each commit should describe what it's doing.

### DIFF
--- a/pkg/smokescreen/conntrack/instrumented_conn_test.go
+++ b/pkg/smokescreen/conntrack/instrumented_conn_test.go
@@ -1,10 +1,10 @@
 package conntrack
 
 import (
+	"fmt"
 	"io"
 	"log"
 	"net"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkg/smokescreen/conntrack/instrumented_conn_test.go
```diff
@@ -27,26 +27,33 @@ func TestInstrumentedConnByteCounting(t *testing.T) {
 	tr := NewTestTracker(0)
 	sent := []byte("X-Smokescreen-Test")
 
-	var wg sync.WaitGroup
-	wg.Add(2)
-
+	writerErrChan := make(chan error, 1)
 	go func() {
-		defer wg.Done()
+		defer close(writerErrChan)
```
cds2-stripe, line 32: This is a much better pattern which makes it easier to pass back contextual errors.

pkg/smokescreen/smokescreen.go
```diff
@@ -521,7 +521,7 @@ func logProxy(config *Config, pctx *goproxy.ProxyCtx) {
 	if pctx.Req.TLS != nil && len(pctx.Req.TLS.PeerCertificates) > 0 {
 		fields["src_host_common_name"] = pctx.Req.TLS.PeerCertificates[0].Subject.CommonName
 		var ouEntries = pctx.Req.TLS.PeerCertificates[0].Subject.OrganizationalUnit
-		if ouEntries != nil && len(ouEntries) > 0 {
+		if len(ouEntries) > 0 {
```
cds2-stripe, line 524: `ouEntries` is a slice of strings, so this check is sufficient.

cds2-stripe (approved):
This is great. Appreciate the cleanup!

Issue #131, opened 2020-12-08
Support upstream proxies


eastebry:
Hello! I'm not sure if this project is still maintained, or if this is a feature you are interested in supporting within Smokescreen.

I'd like to be able to configure smokescreen so that it forwards (allowed) requests to an upstream proxy. Go programs generally respect the `http_proxy`, `https_proxy`, `no_proxy` environment variables. Running smokescreen with this invocation almost works:
```
$ HTTP_PROXY=http://10.0.0.1:9999 HTTPS_PROXY=http://10.0.0.1:9999 ./smokescreen --allow-range 10.0.0.1/32
```
HTTP connections work, CONNECT requests will give this error:
```
 level":"error","msg":"Failed to connect to remote host: parse \"192.168.0.12:9999\": first path segment in URL cannot contain colon"
```

I've traced this error back to stripe/goproxy, and opened a PR: https://github.com/stripe/goproxy/pull/18. `CONNECT` requests should work after merging this PR.

Thanks! If upstream proxies is not a feature you want to support, no worries, I'll look elsewhere.

mattm-stripe:
fixed in #132

PR #132, opened 2020-12-15
Bumping goproxy version to include support for upstream proxies

eastebry:
Bumping the goproxy version, which will include these changes: https://github.com/stripe/goproxy/pull/18, resolving #131


mattm-stripe (changes requested):
Please also `go mod vendor`

mattm-stripe (approved):

PR #135, opened 2021-02-03
Improve smokescreen log messages

rwg-stripe:
r? @cds2-stripe 
cc @stripe/platform-security 

This PR makes changes related to how we log events in smokescreen:

- It sets logrus's default timestamp format (for its `time` field) to include nanoseconds, in order to match other fields that are timestamps.
- It changes `src` and `dst` to `inbound` and `outbound` in field names, as appropriate.  The fields that contain IP addresses/ports will be consistently named `(inbound|outbound)_(local|remote)_addr`, which hopefully makes it clearer what's being referred to.
- It renames the CN/OU fields from `src_host_common_name` and `src_host_organization_unit` to `inbound_remote_x509_cn` and `inbound_remote_x509_ou`, to further remove `src` in field names.
- It updates the smokescreen context's logger with certain fields as soon as it learns the information for those fields.  The biggest thing here is that it'll thread `role`, `project`, and X.509 CN/OU fields through to later log messages instead of only including them in the decision message.

### Known deficiencies

- My goal was to make error messages for TCP connections that failed (timed out, connection refused, etc.) include the failed connections' `outbound_local_addr`, to aid in debugging network issues.  Unfortunately, Go's `net` package doesn't make it easy to obtain the local IP address and port number from a failed TCP connection, and I can't figure out any way to do it short of reimplementing a big chunk of the `net` package.
- Some errors emitted by goproxy do not use the logger in the smokescreen context, so they'll still be missing all kinds of useful information.  This PR doesn't attempt to improve our story here.  (If I had the time to fix it, I think I'd instead spend it on getting rid of goproxy entirely.)

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -372,11 +372,11 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
 
 	logger := cfg.Log.WithFields(logrus.Fields{
 		"id":                  xid.New().String(),
+		"inbound_remote_addr": req.RemoteAddr,
 		"proxy_type":          proxyType,
 		"requested_host":      req.Host,
 		"start_time":          start.UTC(),
 		"trace_id":            req.Header.Get(traceHeader),
-		"inbound_remote_addr": req.RemoteAddr,
 	})
 
 	return &smokescreenContext{
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -263,6 +263,23 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 	}
 	sctx.cfg.StatsdClient.Incr("cn.atpt.success.total", []string{}, 1)
 
+	if conn != nil {
+		fields := logrus.Fields{}
+
+		if addr := conn.LocalAddr(); addr != nil {
+			fields["outbound_local_addr"] = addr.String()
+		}
+
+		if addr := conn.RemoteAddr(); addr != nil {
+			fields["outbound_remote_addr"] = addr.String()
+		}
+
+		if len(fields) > 0 {
+			// add the above fields to all future log messages sent using this smokescreen context's logger
+			sctx.logger = sctx.logger.WithFields(fields)
+		}
+	}
+
 	// Only wrap CONNECT conns with an InstrumentedConn. Connections used for traditional HTTP proxy
 	// requests are pooled and reused by net.Transport.
 	if sctx.proxyType == connectProxy {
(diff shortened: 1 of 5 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -497,39 +516,34 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
...
+	sctx.logger = sctx.logger.WithFields(fields)
+
+	// start a new set of fields used only in this log message
+	fields = logrus.Fields{}
+
+	if pctx.Resp != nil {
+		fields["content_length"] = pctx.Resp.ContentLength
+	}
+
+	if sctx.decision != nil {
```
cds2-stripe, line 546: Could we roll this up into the block above which checks for `sctx.decision != nil`?

pkg/smokescreen/smokescreen.go
```diff
@@ -351,13 +369,14 @@ func configureTransport(tr *http.Transport, cfg *Config) {
...
-		"requested_host": req.Host,
-		"source_addr":    req.RemoteAddr,
-		"start_time":     start.UTC(),
-		"trace_id":       req.Header.Get(traceHeader),
+		"id":                  xid.New().String(),
+		"proxy_type":          proxyType,
+		"requested_host":      req.Host,
+		"start_time":          start.UTC(),
+		"trace_id":            req.Header.Get(traceHeader),
+		"inbound_remote_addr": req.RemoteAddr,
```
cds2-stripe, line 379: Nit: Could we keep these field names in order?

cds2-stripe (changes requested):
This all looks good to me! Just a few small nits.

I agree that the new field names are much easier to understand. And as you said, the time spent figuring out how to reliability shuttle errors back from `goproxy` is better spent removing the dependency altogether. We had to add a layers of custom callbacks just to share a minimal amount of state.

pkg/smokescreen/smokescreen.go
rwg-stripe: If we do that, then those three fields (`decision_reason`, `enforce_would_deny`, and `allow`) would get added to all future log messages written by this smokescreen context's `logger` instead of just this `CANONICAL-PROXY-DECISION` message.

cds2-stripe: Yeah, sorry about that. My eyes skipped over that being reassigned to `sctx.logger`.

pkg/smokescreen/smokescreen.go
rwg-stripe: Reordered in eeaa285d29b702174b6eca5426843fd48727396e.

cds2-stripe (approved):

PR #137, opened 2021-03-05
Adding the ability to define a custom reject response handler

eastebry:
I think it would be nice to be able to customize reject responses. This tiny change adds a RejectResponseHandler to the config struct, allowing you to show custom error responses

It can be used like this:

```
func main() {
	conf, _ := cmd.NewConfiguration(nil, nil)

	conf.RejectResponseHandler = func(resp *http.Response) {
		if resp.StatusCode == http.StatusProxyAuthRequired || resp.StatusCode == http.StatusForbidden {
			resp.Header.Set("X-Custom-Header", "some customer header")
		}
	}
```

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -71,11 +71,6 @@ type Config struct {
 
 	// Customer handler to allow clients to modify reject responses
 	RejectResponseHandler func(*http.Response)
-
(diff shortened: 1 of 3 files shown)


mattm-stripe (approved):
I think this looks fine, though I wonder if we shouldn't allow the entire `rejectResponse` function to be pluggable. I think that's a bit bigger of a change though.

PR #140, opened 2021-04-05
Allow private ranges default option

eastebry:
👋 Hi! We are planning on using smokescreen, but need the option to configure smokescreen in a "blocklist" mode, where private IP ranges are allowed by default, unless they are explicitly blocked. 

For example, we need to allow access to all of `172.16.0.0/12`, but block `172.17.0.0/16`. This is not currently possible with smokescreen, since the [allow-list applies before the block-list](https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/smokescreen.go#L151-L154). `./smokescreen --allow-range 172.16.0.0/12 --deny-range 172.17.0.0/16` will not block the range `172.17.0.0/16`.

The simplest solution is to configure smokescreen in "blocklist mode", . This change introduces the option to pass an `--unsafe-allow-private-ranges` flag which instructs smokescreen to allow private ranges by default, but still respect the `--deny-range` option. (_Note: this only inverts behavior for RFC 1918 ranges. Stuff like 169.254.269.254 will still be blocked by default. I'm open to changing that if you prefer!_)

I'm not totally sure if this is something you want to include in smokescreen. This is definitely the easiest option for us, but I can understand the argument that this behavior would be better in a smokescreen fork. Alternatively, we could do some CIDR math to compute exactly which ranges should be allowed/denied. That's potentially doable, but would get a bit complicated for us (for reasons that are hard to explain :) )

### DIFF
--- a/cmd/smokescreen.go
+++ b/cmd/smokescreen.go
@@ -122,6 +122,10 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
 			Value: "700",
 			Usage: "Set the filemode to `FILE_MODE` on the statistics socket",
 		},
+		cli.BoolFlag{
+			Name:  "unsafe-allow-private-ranges",
+			Usage: "Allow private ip ranges by default",
+		},
 	}
 
 	app.Action = func(c *cli.Context) error {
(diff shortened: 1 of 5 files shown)


eastebry:
Hey Smokescreen maintainers! Gentle bump to see if you have thoughts on this? Incorporating this change would be very nice for us - we wouldn't need to maintain a fork. In the other hand, I understand that this insecure command line option could be a potential footgun for users.

Thanks!

PR #141, opened 2021-04-08
Update documentation to remove mention of `X-Smokescreen-Role`

alexmv:
a94c48a4f4ee removed the default behaviour that checked this header.


alexmv:
Would you instead be amenable to adding a command-line flag which supported the header?  Presumably with documentation caveat'ing the same as https://github.com/stripe/smokescreen/blob/a94c48a4f4ee2d237dc88a9d3a641b9d7e5dc44c/main.go#L22-L23 did?

alexmv:
Friendly nudge on this?

alexmv:
https://github.com/alexmv/smokescreen/tree/flag is a very quick stab at bringing this back, protected by a command-line flag.

PR #143, opened 2021-04-15
Support adding persistent tags to smokescreen metrics

cds2-stripe:
This PR adds support for adding persistent tags to known Smokescreen metrics. This is handled by a new `MetricsClient` struct which wraps the tags and replaces our use of `statsd.Client` with `statsd.ClientInterface`. `MetricsClient` is just a very thin wrapper around the functions of `statsd.ClientInterface` that we utilize.

`MetricsClient` that are initialized without an `addr` parameter use a `statsd.NoOpClient` under the hood. Previously we would just let this persist as nil `*statsd.Client` which I considered to be more of a footgun.

I made the decision to only allow adding persistent tags to known smokescreen metrics. This prevents attempting to tag a metric that doesn't exist, and if you've hooked into smokescreen to emit your own metrics you're already able to provide additional tags to the `MetricsClient`. I also didn't const-ify the metric names, so let me know if you feel strongly about it.

I wrote some very basic unit tests and deployed this to QA internally to verify adding tags worked and existing metrics were emitted as expected.

cc @stripe/platform-security 
r? @mattm-stripe


pkg/smokescreen/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (mc *MetricsClient) Timing(metric string, d time.Duration, rate float64) error {
+	mTags := mc.GetMetricTags(metric)
+	return mc.StatsdClient.Timing(metric, d, mTags, rate)
+}
+
+func (mc *MetricsClient) TimingWithTags(metric string, d time.Duration, rate float64, tags []string) error {
+	mTags := mc.GetMetricTags(metric)
+	tags = append(tags, mTags...)
```
mattm-stripe, line 97: pkg/smokescreen/metrics.go:97:2: ineffectual assignment to `tags` (ineffassign)
	tags = append(tags, mTags...)

pkg/smokescreen/metrics.go
```diff
@@ -0,0 +1,99 @@
+package smokescreen
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/DataDog/datadog-go/statsd"
+)
+
+var metrics = map[string][]string{
```
mattm-stripe, line 10: Please add a comment explaining what this global is for

pkg/smokescreen/metrics.go
```diff
@@ -0,0 +1,99 @@
...
+		additionalTags: make(map[string][]string),
+		StatsdClient:   &statsd.NoOpClient{},
+	}
+}
+
+// AddMetricTags associates the provided tags slice with a given metric. The metric must be present
+// in the metrics slice.
+func (mc *MetricsClient) AddMetricTags(metric string, mTags []string) error {
+	if tags, ok := metrics[metric]; ok {
+		metrics[metric] = append(tags, mTags...)
```
mattm-stripe, line 65: Are there any concurrency concerns around modifying a global?  Maybe this should just be a field inside the MetricsClient instead, along with possibly a lock?

mattm-stripe (changes requested):
A few relatively minor concerns but lgtm otherwise

pkg/smokescreen/metrics.go
cds2-stripe: Nice catch! I'm surprised my VSCode linter didn't yell at me 🤔

pkg/smokescreen/metrics.go
cds2-stripe: Yeah totally agree about moving this inside the `MetricsClient` struct. The intent is that persistent metrics should only be added when the configuration is being initialized, and that `MetricsClient` isn't intended to be thread safe.

Given how concurrent we run smokescreen, I'd be concerned about lock contention if we dropped a mutex in there and would even prefer to avoid an RWLock if possible.

I added comments clarifying the thread safety and intent around MetricsClient.

cds2-stripe: I also added a guard that gets set when smokescreen starts to prevent adding new metrics tags.

cds2-stripe:
ptal? @mattm-stripe

mattm-stripe (approved):
thanks! looks good.

PR #144, opened 2021-04-20
allow ipv4 only dns lookups

cds2-stripe:
This PR allows Smokescreen to be configured to only perform IPv4 lookups. `LookupIPAddr` looks up both v4 and v6 addresses, which doubles the total number of DNS queries to the configured resolver. Go 1.15 introduced a new function, [LookupIP](https://golang.org/pkg/net/#Resolver.LookupIP), which only performs v4 lookups. We will continue to default to the old behavior but allow clients to configure Smokescreen to only perform v4 lookups.

r? @mattm-stripe 
cc @stripe/platform-security

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -55,6 +55,9 @@ type Config struct {
 	Healthcheck                  http.Handler // User defined http.Handler for optional requests to a /healthcheck endpoint
 	ShuttingDown                 atomic.Value // Stores a boolean value indicating whether the proxy is actively shutting down
 
+	// Only resolve ipv4 addrs during DNS lookups
+	IPv4OnlyLookups bool
+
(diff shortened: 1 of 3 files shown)


mattm-stripe (changes requested):
LookupIPAddr is just LookupIP with the second argument set to "ip" -- I think it makes more sense to remove the branching behavior between LookupIP/LookupIPAddr, and instead keep a `config.Network` or similar variable.

As an alternative approach, we already have Config.Resolver pluggable -- Maybe we could add a small interface with LookupIPAddr, and a thin wrapper implementation using LookupIP, but with the network type configured internally.  I think that might be getting a bit too abstract, though, and could lead to some potentially tricky interactions -- config.SetResolverAddresses as-is just overwrites resolver, so we'd need to make sure they play together in an understandable way.

mattm-stripe (approved):

PR #145, opened 2021-06-24
Disallow wildcard ACL entries that match all hostnames

rwg-stripe:
r? @cds2-stripe 
cc @stripe/platform-security 

Explicitly disallow adding ACL entries with a domain of `*` or `*.` and update the tests to ensure these are rejected.  (`*` is already rejected by the tests, but `*.` isn't.)

Example test failure (after commenting out a check in `acl.go`):

```
--- FAIL: TestACLAddInvalidDomain (0.00s)
    acl_test.go:237: 
        	Error Trace:	acl_test.go:237
        	Error:      	An error is expected but got nil.
        	Test:       	TestACLAddInvalidDomain
        	Messages:   	did not reject invalid domain "*."
```

### DIFF
--- a/pkg/smokescreen/acl/v1/acl.go
+++ b/pkg/smokescreen/acl/v1/acl.go
@@ -181,6 +181,10 @@ func (acl *ACL) ValidateDomains(domains []string) error {
 			return fmt.Errorf("glob cannot be empty")
 		}
 
+		if d == "*" || d == "*." {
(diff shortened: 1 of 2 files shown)


cds2-stripe (approved):
LGTM. Thanks for updating the tests!

ransford-stripe:
Rolled into #151.

PR #150, opened 2021-12-15
add dns and tcp timing information to logs

cds2-stripe:
This PR adds TCP connection establishment and DNS lookup timing information to smokescreen's logs. This is being done to provide more granular connection information to help debug cases where a downstream service has a tight timeout window. Currently when a connection is timed out by the requestor it can be difficult to understand where time was spent in the connection initialization process.

`dns_lookup_time` will be logged with every CANONICAL-PROXY-DECISION log line. As we pre-resolve the requested host, we're able to attach this to the decision log. This field represents the time spent resolving the requested hostname.

`tcp_establish_time` is the amount of time spent establishing the TCP connection. Because of the asynchronous way `goproxy` invokes our custom `net.Transport` we can only add this field to the CANONICAL-PROXY-CN-CLOSE log line. `goproxy` requires us to decide whether to proxy the request prior to establishing the TCP connection. This means that we log the proxy decision before the TCP connection is established.

I've deployed this internally and verified the fields are being populated as expected.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -54,6 +54,9 @@ type smokescreenContext struct {
 	proxyType     string
 	logger        *logrus.Entry
 	requestedHost string
+
+	// Time spent resolving the requested hostname
+	lookupTime time.Duration
 }
 
 // ExitStatus is used to log Smokescreen's connection status at shutdown time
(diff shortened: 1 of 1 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -240,20 +243,25 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
...
+	start := time.Now()
 	if sctx.cfg.ProxyDialTimeout == nil {
 		conn, err = net.DialTimeout(network, d.resolvedAddr.String(), sctx.cfg.ConnectTimeout)
 	} else {
 		conn, err = sctx.cfg.ProxyDialTimeout(ctx, network, d.resolvedAddr.String(), sctx.cfg.ConnectTimeout)
 	}
+	connTime := time.Since(start)
+
+	fields := logrus.Fields{
+		"conn_establish_time": connTime.String(),
```
stroopc-stripe, line 259: I noticed this filed and the `dns_lookup_time` field aren't consistent in their unit of time they're logged in. Sometimes it's in `ms` and sometimes in microseconds (and I assume sometimes seconds if it takes long enough). Should we standardize the time before logging it for ease of sorting/searching?

stroopc-stripe (changes requested):
mostly lgtm, one question before approving

pkg/smokescreen/smokescreen.go
```diff
@@ -256,7 +256,7 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 	connTime := time.Since(start)
 
 	fields := logrus.Fields{
-		"conn_establish_time": connTime.String(),
+		"conn_establish_time_ms": connTime.Milliseconds(),
```
stroopc-stripe, line 259: mm, [it looks like this](https://pkg.go.dev/time#Duration.Milliseconds) will round everything to the nearest millisecond which means anything less than 1 millisecond will be logged as 0ms, which may be fine, if we don't care about that fine grain of timing on sub 1 millisecond requests.

stroopc-stripe (changes requested):

pkg/smokescreen/smokescreen.go
```diff
@@ -545,6 +545,9 @@ func logProxy(config *Config, pctx *goproxy.ProxyCtx) {
 
 	// start a new set of fields used only in this log message
 	fields = logrus.Fields{}
+
+	// If a lookup takes less than 1ms it will be rounded down to zero. This can separated from
```
stroopc-stripe, line 549:
```suggestion
	// If a lookup takes less than 1ms it will be rounded down to zero. This can be separated from
```

stroopc-stripe (approved):
lgtm, thanks for sending this in!

stroopc-stripe (approved):

stroopc-stripe (approved):

PR #153, opened 2022-03-09
bump goproxy to 3f1dfba6d1a4747c78fee7069e2f28fd6b703917

cds2-stripe:
This bumps the version of `stripe/goproxy` to the latest master. Specifically it brings in two commits:
* [Limit response size from upstream proxies](https://github.com/stripe/goproxy/pull/20)
* [Add KeepAcceptEncoding option to keep the Accept-Encoding header intact ](https://github.com/stripe/goproxy/pull/19)

I also fixed an integration test that changed since the last PR. You can verify the new behavior is correct manually with:
```
curl -I https://api.stripe.com
HTTP/2 404
```

r? @jjiang-stripe

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -1,3 +1,4 @@
+//go:build !nointegration
 // +build !nointegration
 
 package cmd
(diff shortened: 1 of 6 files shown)


jjiang-stripe (approved):
lgtm! I verified the diff is accurate and the integration test change is correct

PR #156, opened 2022-04-07
Go 1.18

cds2-stripe:
r? @jjiang-stripe 

This bumps the supported Go version to 1.18 and drops support for Go 1.16. I also fixed a test that was causing failures in https://github.com/stripe/smokescreen/pull/155.


jjiang-stripe (approved):
lgtm! the test fix looks reasonable based on the error in #155

PR #157, opened 2022-04-20
Use tags to track successful/failed connection attempt metrics

jjiang-stripe:
r? @cds2-stripe 

Defining these as two separate metrics means it's difficult to get an accurate success rate because metrics occasionally get reported into separate windows. By using tags instead, we can properly measure success/failure rates. This PR does exactly that! 

I deployed this internally and verified that metrics continue to be reported as expected, with the appropriate tag value.


cds2-stripe (approved):
This LGTM! Thanks for fixing this.

jjiang-stripe:
r-codeowners?

PR #158, opened 2022-04-28
Fix hostname parsing for hosts with square brackets

jjiang-stripe:
r? @cds2-stripe

This fixes up our handling for hostnames that contain square brackets, like `[stripe.com]`. We check against our ACL using `[stripe.com]` instead of `stripe.com` and then make a DNS query + request for `stripe.com`, which is confusing and inconsistent (the request ends up being malformed, but it's still bizarre). I've done a bit of hackery to get from `[stripe.com]` -> `stripe.com:80` as cleanly as possible without breaking things for IPv6 addresses. I've stared at this code possibly a little _too_ long so it's quite likely I've overengineered a solution here and I'm very open to alternatives 😅

I added a unit test to confirm the behaviour, and will deploy this internally to verify existing use cases still work as intended.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -414,6 +414,36 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
 	}
 }
 
+func stripSquareBrackets(host string) string {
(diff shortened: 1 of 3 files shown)


cds2-stripe (approved):
This LGTM! I tested this internally as well to verify it didn't impact existing ACL rules.

PR #160, opened 2022-06-11
Use Go 1.17 IP.IsPrivate() function.

### LINKED ISSUE #148: Golang 1.17 adds a new IP.IsPrivate function we could use
https://pkg.go.dev/net#IP.IsPrivate

This function could be used instead of our own PrivateRuleRanges, 

https://github.com/stripe/smokescreen/blob/dc403015f563eadc556a61870c6ad327688abe88/pkg/smokescreen/smokescreen.go#L155
https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/constants.go#L20-L25
https://github.com/stripe/smokescreen/blob/master/pkg/smokescreen/constants.go#L38-L45

As Smokescreen supports the two most recent Go releases, we should wait until 1.18 is released so that we can support both 1.17 and 1.18

alexmv:
Fixes #148.

### DIFF
--- a/pkg/smokescreen/constants.go
+++ b/pkg/smokescreen/constants.go
@@ -1,7 +1,6 @@
 package smokescreen
 
 import (
-	"net"
 	"regexp"
 )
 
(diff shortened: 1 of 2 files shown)


alexmv:
See #162 for the fix for the failing tests.

jjiang-stripe (approved):
thanks for keeping us up-to-date with new standard library functions!

PR #161, opened 2022-06-11
Accept a flag which parses the role from an HTTP header.

alexmv:
a94c48a4f4ee removed this default behaviour; add it back, though
protected by a flag.

---

This came up in #141, and we'd find it useful behaviour to have available.

"import `smokescreen` and override `conf.RoleFromRequest`" is a fine response to this, but I think it's a useful low-stakes way to configure roles, and provide out-of-the-box.


alexmv:
See #162 for the fix for the failing tests.

PR #162, opened 2022-06-11
tests: Do not assume example.com is a public IP.

alexmv:
Tests are failing in Github CI because example.com resolves to a
private IP address.


alexmv:
Gentle nudge on this?

jjiang-stripe:
hmm it looks like maybe this was just a transient issue? I re-ran the tests on https://github.com/stripe/smokescreen/pull/160 and https://github.com/stripe/smokescreen/pull/161 and it looks like they pass now

alexmv:
Huh, weird!  Makes sense to drop, then.

PR #163, opened 2022-06-24
Strip all square brackets

jjiang-stripe:
r? @cds2-stripe 

As a follow up to https://github.com/stripe/smokescreen/pull/158, if you supply a hostname with _two_ sets of square brackets (e.g. [[example.com]]), it ends up only stripping one set ([example.com]) and we end up falling into the old code path that allowed the request. Converting to a loop _should_ fix things. I also added a bit of error handling in the ACL matching code that would panic if you passed something like `[[example.com]]:80` because `net.SplitHostPort` would fail, and we'd end up passing `[[[example.com]]:80]:80`. Given that the hostname is malformed anyway, I figured it wasn't worth trying to parse it into something manageable, and we should just deny and return an error.

I've added several tests to confirm the behaviour. I also ran this and tested an example set up to confirm this now works.

### DIFF
--- a/pkg/smokescreen/smokescreen_test.go
+++ b/pkg/smokescreen/smokescreen_test.go
@@ -443,11 +443,18 @@ func TestInvalidHost(t *testing.T) {
 }
 
 var hostSquareBracketsCases = []struct {
-	scheme    string
-	proxyType string
+	scheme         string
+	proxyType      string
+	hostname       string
+	decisionReason string
 }{
-	{"http", "http"},
-	{"https", "connect"},
+	{"http", "http", "[stripe.com]", "host matched rule in global deny list"},
+	{"https", "connect", "[stripe.com]", "host matched rule in global deny list"},
+	{"http", "http", "[[stripe.com]]", "host matched rule in global deny list"},
+	{"https", "connect", "[[stripe.com]]", "host matched rule in global deny list"},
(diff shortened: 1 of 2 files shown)


pkg/smokescreen/smokescreen_test.go
```diff
@@ -443,11 +443,18 @@ func TestInvalidHost(t *testing.T) {
...
+	decisionReason string
 }{
-	{"http", "http"},
-	{"https", "connect"},
+	{"http", "http", "[stripe.com]", "host matched rule in global deny list"},
+	{"https", "connect", "[stripe.com]", "host matched rule in global deny list"},
+	{"http", "http", "[[stripe.com]]", "host matched rule in global deny list"},
+	{"https", "connect", "[[stripe.com]]", "host matched rule in global deny list"},
+	{"http", "http", "[[[stripe.com]]]", "host matched rule in global deny list"},
+	{"https", "connect", "[[[stripe.com]]]", "Destination host cannot be determined"},
```
jjiang-stripe, line 456: I don't know why this happens, but I assume it's got something to do with how Go constructs the CONNECT request. These should both be denied anyway, so I wasn't too concerned about the specific error

jjiang-stripe:
I deployed this internally and didn't see any issues either!

cds2-stripe (approved):
This LGTM! I think failing faster here against unexpected hostname formatting is a better approach.

PR #164, opened 2022-07-12
Improve address normalization

lewandowski-stripe:
- Use the same address for connecting to destination and for ACL checks
- Support IPv6 addresses
- Remove regexp-based domain validation
- Enforce a standardized form of IP addresses and DNS names
- Add tests for `normalizeHost()`

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -526,7 +526,7 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 	})
 
 	// Handle CONNECT proxy to TLS & other TCP protocols destination
-	proxy.OnRequest().HandleConnectFunc(func(host string, pctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
+	proxy.OnRequest().HandleConnectFunc(func(_ string, pctx *goproxy.ProxyCtx) (*goproxy.ConnectAction, string) {
 		pctx.UserData = newContext(config, connectProxy, pctx.Req)
 		pctx.HTTPErrorHandler = HTTPErrorHandler
 
@@ -535,12 +535,12 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 		defer logProxy(config, pctx)
 		defer pctx.Req.Header.Del(traceHeader)
 
-		err := handleConnect(config, pctx)
+		destination, err := handleConnect(config, pctx)
 		if err != nil {
 			pctx.Resp = rejectResponse(pctx, err)
 			return goproxy.RejectConnect, ""
 		}
-		return goproxy.OkConnect, host
+		return goproxy.OkConnect, destination
 	})
 
 	// Strangely, goproxy can invoke this same function twice for a single HTTP request.
@@ -634,24 +634,24 @@ func logProxy(config *Config, pctx *goproxy.ProxyCtx) {
 	logMethod(CanonicalProxyDecision)
 }
 
-func handleConnect(config *Config, pctx *goproxy.ProxyCtx) error {
+func handleConnect(config *Config, pctx *goproxy.ProxyCtx) (string, error) {
 	sctx := pctx.UserData.(*smokescreenContext)
 
 	// Check if requesting role is allowed to talk to remote
 	remoteHost, remotePort, err := normalizeHost(pctx.Req.Host, pctx.Req.URL.Scheme)
 	if err != nil {
 		pctx.Error = err
-		return pctx.Error
+		return "", pctx.Error
 	}
 	sctx.decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, pctx.Req, remoteHost, remotePort)
 	if pctx.Error != nil {
-		return pctx.Error
+		return "", pctx.Error
 	}
 	if !sctx.decision.allow {
-		return denyError{errors.New(sctx.decision.reason)}
+		return "", denyError{errors.New(sctx.decision.reason)}
 	}
 
-	return nil
+	return net.JoinHostPort(remoteHost, strconv.Itoa(remotePort)), nil
 }
 
 func findListener(ip string, defaultPort uint16) (net.Listener, error) {
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -452,7 +452,7 @@ func normalizeHost(hostPort, scheme string, forceFQDN bool) (string, int, error)
 		if err != nil {
 			return "", noPort, fmt.Errorf("invalid port number %#v: %v", port, err)
 		}
-		if port < portMin && port > portMax {
+		if port < portMin || port > portMax {
 			return "", noPort, fmt.Errorf("invalid port number %#v: must be between %d and %d", port, portMin, portMax)
 		}
 	}
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -618,24 +656,24 @@ func logProxy(config *Config, pctx *goproxy.ProxyCtx) {
...
-		pctx.Error = err
-		return pctx.Error
+		pctx.Error = denyError{err}
+		return "", pctx.Error
 	}
-	sctx.decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, pctx.Req, remoteHost)
+	sctx.decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, pctx.Req, remoteHost, remotePort)
 	if pctx.Error != nil {
-		return pctx.Error
+		return "", pctx.Error
```
jjiang-stripe, line 670: should this one also be wrapped?

```suggestion
		return "", denyError{pctx.Error}
```

pkg/smokescreen/smokescreen.go
```diff
@@ -895,14 +934,14 @@ func checkACLsForRequest(config *Config, req *http.Request, outboundHost string)
 
 	decision.role = role
 
-	submatch := hostExtractRE.FindStringSubmatch(outboundHost)
-	if len(submatch) < 2 {
+	// This host validation prevents IPv6 addresses from being used as destinations.
+	// Added for backwards compatibility.
+	if strings.ContainsAny(host, ":") {
```
jjiang-stripe, line 939: is this just because the regex failed on IPv6 addresses before?

pkg/smokescreen/smokescreen.go
```diff
@@ -852,13 +890,14 @@ func getRole(config *Config, req *http.Request) (string, error) {
 	}
 }
 
-func checkIfRequestShouldBeProxied(config *Config, req *http.Request, outboundHost string) (*aclDecision, time.Duration, error) {
-	decision := checkACLsForRequest(config, req, outboundHost)
+func checkIfRequestShouldBeProxied(config *Config, req *http.Request, host string, port int) (*aclDecision, time.Duration, error) {
+	decision := checkACLsForRequest(config, req, host, port)
```
jjiang-stripe, line 894: ACL checks will use the normalized host

pkg/smokescreen/smokescreen.go
```diff
@@ -852,13 +890,14 @@ func getRole(config *Config, req *http.Request) (string, error) {
...
 
 	var lookupTime time.Duration
 	if decision.allow {
 		start := time.Now()
-		resolved, reason, err := safeResolve(config, "tcp", outboundHost)
+		hostPort := net.JoinHostPort(host, strconv.Itoa(port))
+		resolved, reason, err := safeResolve(config, "tcp", hostPort)
```
jjiang-stripe, line 900: DNS resolution also uses the normalized host

pkg/smokescreen/smokescreen.go
```diff
@@ -519,12 +557,12 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
...
 		defer pctx.Req.Header.Del(traceHeader)
 
-		err := handleConnect(config, pctx)
+		destination, err := handleConnect(config, pctx)
 		if err != nil {
 			pctx.Resp = rejectResponse(pctx, err)
 			return goproxy.RejectConnect, ""
 		}
-		return goproxy.OkConnect, host
+		return goproxy.OkConnect, destination
```
jjiang-stripe, line 565: this is kind of interesting: we're now using the normalized host name instead of the one passed on the CONNECT request (which varies based on http clients). I think this makes sense though!

jjiang-stripe (changes requested):
I think this looks good and is more consistent compared to the hacks I shoved in 😅

I left a couple minor comments! we should also deploy this internally to make sure there aren't any regressions (I can help with this!)

pkg/smokescreen/smokescreen.go
```diff
@@ -414,32 +416,68 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
...
+// standardized IP address representation), as well as the port number.
+func normalizeHost(hostPort, scheme string, forceFQDN bool) (string, int, error) {
+	var err error
+	const noPort = -1
+	host, port := hostPort, noPort
+
+	// net.SplitHostPort() doesn't handle bare IPv6 addresses well so
+	// handle that case first.
+	if ip := net.ParseIP(hostPort); ip != nil && ip.To4() == nil {
+		host = ip.String()
```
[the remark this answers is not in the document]
lewandowski-stripe: Correct. Is the comment above sufficient or should I improve it?

pkg/smokescreen/smokescreen.go
```diff
@@ -414,32 +416,68 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
...
+	} else if hasPort(hostPort) {
+		// Extract host and port if both are provided.
+		var portString string
+		host, portString, err = net.SplitHostPort(hostPort)
+		if err != nil {
+			return "", noPort, fmt.Errorf("unable to parse host: %v", err)
+		}
```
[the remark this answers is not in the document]
lewandowski-stripe: If we reached this code `hasPort()` returned true. So either it's a bare IPv6 address (already ruled out in the opening if statement) or the `hostPort` is indeed malformed (since it was supposed to contain a port but it does not).

pkg/smokescreen/smokescreen.go
```diff
@@ -414,32 +416,68 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
...
+		if port < portMin && port > portMax {
+			return "", noPort, fmt.Errorf("invalid port number %#v: must be between %d and %d", port, portMin, portMax)
 		}
 	}
-	return net.JoinHostPort(hostname, port), nil
+
+	if port == noPort {
+		// Port was not provided so try to determine it based on scheme.
+		port, err = net.LookupPort("tcp", scheme)
```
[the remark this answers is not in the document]
lewandowski-stripe: Yes. I think it's preferable to fail explicitly at this stage then assign a meaning to port 0.

pkg/smokescreen/smokescreen.go
```diff
@@ -414,32 +416,68 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
...
+		if err != nil {
+			return "", noPort, fmt.Errorf("unable to determine port for %v", scheme)
+		}
+	}
+
+	if ip := net.ParseIP(host); ip != nil {
+		host = ip.String()
+	} else {
```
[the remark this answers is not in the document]
lewandowski-stripe: Not really. IP addresses can have multiple representations (e.g., `2001:DB8::` and `2001:db8::` represent the same address but they are different strings). We should ensure that we are consistent about which format we use.

pkg/smokescreen/smokescreen.go
lewandowski-stripe: It probably should. It wasn't so I didn't touch it. ;)

pkg/smokescreen/smokescreen.go
lewandowski-stripe: Yes. I hope to remove it in one of subsequent PRs as I hope to implement the IPv6 support. But for now I'd like to keep the current behaviour.

pkg/smokescreen/smokescreen.go
lewandowski-stripe: Yes, WAI.

pkg/smokescreen/smokescreen.go
lewandowski-stripe: We should use the same host that we used for ACL checking. This way, even if someone is able to workaround host normalization, they only should be able to access destinations they were already authorized for.

pkg/smokescreen/smokescreen.go
lewandowski-stripe: Correct!

pkg/smokescreen/smokescreen.go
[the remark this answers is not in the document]
jjiang-stripe: I think it makes sense! I'm mostly just papertrailing comments for myself so I remember that I understand what this is doing 😅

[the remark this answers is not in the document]
lewandowski-stripe: The code should be self explanatory so the places where you left comments are valuable to me as indicators where comments are missing or code should be made clearer. Thank you!

lewandowski-stripe:
> I think this looks good and is more consistent compared to the hacks I shoved in 😅

Thank you! For what is worth, you addressed the problem quickly so I am glad that you have added these "hacks". ;)

> I left a couple minor comments! we should also deploy this internally to make sure there aren't any regressions (I can help with this!)

Thank you! That'd be very much appreciated!

jjiang-stripe (commented):
this looks good! I'll make a TODO for us to test this out internally first before merging, and I'll come back to approve this once we do!

jjiang-stripe (dismissed):
I deployed this internally to our QA environment and didn't see any DNS resolution failures or unexpected ACL denials, so I think this is good to go!

I also pulled this and did a little local repro for the square bracket CVEs we had _just to be super sure_ and they return denials/errors as expected.

lewandowski-stripe:
r?

Thank you @jjiang-stripe!

pkg/smokescreen/smokescreen.go
```diff
@@ -414,32 +416,74 @@ func newContext(cfg *Config, proxyType string, req *http.Request) *smokescreenCo
...
+		var portString string
+		host, portString, err = net.SplitHostPort(hostPort)
+		if err != nil {
+			return "", noPort, fmt.Errorf("unable to parse host: %v", err)
+		}
+		port, err = strconv.Atoi(portString)
+		if err != nil {
+			return "", noPort, fmt.Errorf("invalid port number %#v: %v", port, err)
+		}
+		if port < portMin && port > portMax {
```
lewandowski-stripe: Thank you @cds2-stripe for noticing the if-condition that'd never match!

cds2-stripe (approved):

PR #165, opened 2022-07-12
Use Go-embedded binary version

lewandowski-stripe:
Go 1.18 embeds version control information in binaries: https://tip.golang.org/doc/go1.18#go-version

Deprecate hard coded constants in favour of the new mechanism at the cost of changing version format.

### DIFF
--- a/cmd/smokescreen.go
+++ b/cmd/smokescreen.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/carlmjohnson/versioninfo"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/urfave/cli.v1"
 
@@ -28,7 +29,7 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
 
 	app := cli.NewApp()
 	app.Name = "smokescreen"
-	app.Version = smokescreen.Version()
+	app.Version = versioninfo.Short()
 	app.Usage = "A simple HTTP proxy that prevents SSRF and can restrict destinations"
 	app.ArgsUsage = " " // blank but non-empty to suppress default "[arguments...]"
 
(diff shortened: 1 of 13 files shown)


jjiang-stripe (changes requested):
hmmm so it looks like if I build this with Go 1.18 I get:
```
$ ./smokescreen -v
smokescreen version rev-777820e
```
and 1.17
```
$ ./smokescreen -v
smokescreen version devel
```
It feels like we're losing information in the 1.17 case, so I wonder if we should hold off on this for now? 

We also have a couple security advisories using the old notation: https://github.com/stripe/smokescreen/security/advisories/GHSA-qwrf-gfpj-qvj6 (which I'm now realizing we forgot to bump to 0.0.4 in the code, which this would actually help with 😅). I guess we could change the notation for those, but it becomes less immediately obvious if a version is affected if we only have the git hash instead of something numerical. WDYT?

lewandowski-stripe:
> hmmm so it looks like if I build this with Go 1.18 I get:
> 
> ```
> $ ./smokescreen -v
> smokescreen version rev-777820e
> ```
> 
> and 1.17
> 
> ```
> $ ./smokescreen -v
> smokescreen version devel
> ```

I think if you build it with `go install github.com/stripe/smokescreen@v0.0.4`, the resulting binary should show `v0.0.4` as its version, even for Go 1.17. Of course it's tricky to test because v0.0.4 doesn't contain the proposed change but perhaps we could do it in v0.0.5-alpha or something.

> It feels like we're losing information in the 1.17 case, so I wonder if we should hold off on this for now?

I am happy to wait, there is no rush.

> We also have a couple security advisories using the old notation: [GHSA-qwrf-gfpj-qvj6](https://github.com/stripe/smokescreen/security/advisories/GHSA-qwrf-gfpj-qvj6) (which I'm now realizing we forgot to bump to 0.0.4 in the code, which this would actually help with 😅). I guess we could change the notation for those, but it becomes less immediately obvious if a version is affected if we only have the git hash instead of something numerical. WDYT?

The version should get correctly embedded IIUC.

jjiang-stripe (approved):
ahh got it, this looks good to me!

PR #166, opened 2022-07-22
Use more strict check for CONNECT's host:port

lewandowski-stripe:
https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6 mandates that:

> A client sending a CONNECT request MUST send the authority form of request-target ([Section 5.3 of [RFC7230]](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3)); i.e., the request-target consists of only the host name and port number of the tunnel destination, separated by a colon.

Smokescreen should enforce these rules to lower the risk of parsing malformed requests.

Additionally, upgrade github.com/stretchr/testify to v1.8.0 to make use of the `ErrorsContains()` assertion.


pkg/smokescreen/smokescreen.go
```diff
@@ -421,15 +421,74 @@ func hasPort(s string) bool {
 	return strings.LastIndex(s, "]") < strings.LastIndex(s, ":")
 }
 
-// normalizeHost returns host (as string) and port (as int) derived from the
+func _normalizePort(s string) (port int, err error) {
```
cds2-stripe, line 424: OOC why the leading underscore?

pkg/smokescreen/smokescreen.go
```diff
@@ -421,15 +421,74 @@ func hasPort(s string) bool {
...
+		// (e.g., `2001:DB8::` and `2001:db8::` are the same address).
+		// This function provides a consistent representation.
+		return ip.String(), nil
+	}
+	// If it's not an IP address then it must be a domain name.
+	// Convert it to Punycode it so that we deal only with with ASCII from now on.
+	// This way we can find out whether the domain name is malformed.
+	domain, err := idna.Lookup.ToASCII(s)
+	if err != nil {
+		return "", fmt.Errorf("invalid domain '%v': %v", s, err)
```
cds2-stripe, line 450: FYI you can also use the `%q` format string to wrap a value in quoates.

pkg/smokescreen/smokescreen.go
```diff
@@ -421,15 +421,74 @@ func hasPort(s string) bool {
 	return strings.LastIndex(s, "]") < strings.LastIndex(s, ":")
 }
 
-// normalizeHost returns host (as string) and port (as int) derived from the
+func _normalizePort(s string) (port int, err error) {
+	if strings.HasPrefix(s, "0") && s != "0" {
```
cds2-stripe, line 425: Is the conversion using `strconv.Atoi()` not sufficient? It might obscure some unintended behavior though. Either way a comment about this check might be helpful.

cds2-stripe (changes requested):
Just a couple of small comments, but this looks great!

pkg/smokescreen/smokescreen.go
lewandowski-stripe: Honestly? Because I was lazy and considered them undocumented private functions. I added some comments and removed underscores. Thanks for providing some motivation! :)

pkg/smokescreen/smokescreen.go
lewandowski-stripe: Unfortunately double quotes will get escaped in the log output so it won't look as good.

pkg/smokescreen/smokescreen.go
lewandowski-stripe: I think you are right. That check was superflous and `strconv.Atoi()` will normalize it anyway.

lewandowski-stripe:
> Just a couple of small comments, but this looks great!

Thank you for the comments! I hope I've addressed them and this PR should be ready for merging.

cds2-stripe (approved):

PR #168, opened 2022-08-17
Add feature to track connection success rate 

cmoresco-stripe:
This PR shouldn't cause any change to the way smokescreen operates in any currently-existing configuration. 

It does, however, give users the option to pass their own custom ConnTracker in the smokescreen config, which may include an optional ConnSuccessRateTracker. The latter is a new struct that tracks and periodically the success rate of connection attempts, over a sliding time window of configurable length. If connections to the same host have been attempted more than once in the window, only the most recent attempt is counted (to avoid a single destination host from having an outsized impact.


pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +17,92 @@ type Tracker struct {
...
+	CalculatedAt    time.Time
+	ConnSuccessRate float64
+	TotalConns      int
+}
+
+// StartNewConnSuccessRateTracker creates a new ConnSuccessRateTracker with a specific calculation interval at which
+// ConnSuccessRateStats will be recomputed, and a time window to calculate those statistics over.
+func StartNewConnSuccessRateTracker(calculationInterval time.Duration, calculationWindow time.Duration) *ConnSuccessRateTracker {
+	newSuccessTracker := &ConnSuccessRateTracker{
+		ConnAttempts: cache.New(calculationWindow, time.Second*10),
```
ransford-stripe, line 52: Instead of hardcoding the cleanup interval (the second arg) here, you might use a constant with a comment attached to it. How often cache entries get evicted seems relevant to the math you're doing here, or at least worth explicitly taking into consideration.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +17,92 @@ type Tracker struct {
...
+					succeeded++
+				}
+			}
+			var successRate float64
+			// Avoid divide by zero errors
+			if total == 0 {
+				successRate = float64(100)
+			} else {
+				successRate = (float64(succeeded) / float64(total)) * 100
+			}
```
ransford-stripe, line 71: This calculation looks reasonable to me.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +17,92 @@ type Tracker struct {
...
+		ShuttingDown:       sd,
+		Wg:                 &sync.WaitGroup{},
+		IdleTimeout:        idle,
+		statsc:             statsc,
+		SuccessRateTracker: successRateTracker,
+	}
+}
+
+// RecordAttempt stores the result of the most recent connection attempt for a destination.
+func (tr *Tracker) RecordAttempt(dest string, success bool) error {
```
ransford-stripe, line 93: You might consider giving this function a void return type (i.e., remove `error`) and just quietly returning if `tr.SuccessRateTracker == nil`. That might simplify the code. It looks like you're not doing anything interesting with the error anyway in `recordConnectionAttempt()`.

Since this (and `ReportConnectionSuccessRate()`) seems to be testing for a misconfiguration, would you be able to come up with a way to surface misconfiguration earlier (instead of logging potentially a lot of errors)?

ransford-stripe (changes requested):
I think this approach and implementation seem largely sound! My comments are mostly along the lines of safety and disambiguation.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +17,92 @@ type Tracker struct {
...
 	// A connection is idle if it has been inactive (no bytes in/out) for this
 	// many seconds.
 	IdleTimeout time.Duration
 }
 
-func NewTracker(idle time.Duration, statsc statsd.ClientInterface, logger *logrus.Logger, sd atomic.Value) *Tracker {
+// ConnSuccessRateTracker tracks statistics about the overall success rate
+// of connection attempts over some time interval (which is set in StartNewConnSuccessRateTracker()).
+//
+// It tracks only the *most recently seen* connection to an individual destination host within the configured
```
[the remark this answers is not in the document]
ransford-stripe: Also wondering what normalization you think might be appropriate. If the population of outgoing requests looks like
```
customer-4355a46b19.bighostingsite.com
customer-53c234e5e8.bighostingsite.com
customer-1121cfccd5.bighostingsite.com
customer-7de1555df0.bighostingsite.com
customer-f0b5c2c221.bighostingsite.com
customer-06e9d52c17.bighostingsite.com
customer-10159baf26.bighostingsite.com
customer-aa67a169b0.bighostingsite.com
blorp.lol
customer-917df3320d.bighostingsite.com
```
and `bighostingsite.com` is completely down, it will look like a 90% failure rate.

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -51,3 +51,55 @@ func NewTestTracker(idle time.Duration) *Tracker {
...
+
+			sd := atomic.Value{}
+			sd.Store(false)
+			tracker := NewTracker(time.Second, &statsd.NoOpClient{}, logrus.New(), sd, StartNewConnSuccessRateTracker(500*time.Millisecond, 2*time.Second))
+
+			for _, record := range tc.additions {
+				tracker.RecordAttempt(record.host, record.success)
+			}
+
+			time.Sleep(tc.waitTime)
```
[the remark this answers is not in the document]
cmoresco-stripe: I gave it a try, but since the cache library (https://github.com/patrickmn/go-cache/blob/master/cache_test.go) doesn't support injecting a clock it doesn't give us a great way to test cache expiry.

pkg/smokescreen/conntrack/conn_tracker.go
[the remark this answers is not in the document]
cmoresco-stripe: eTLD+1? There's a nice [go library](https://pkg.go.dev/golang.org/x/net/publicsuffix) for this, I wrote a function to normalize domains below, but let me know if it seems like there are cases I haven't thought of

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +17,92 @@ type Tracker struct {
...
+// impact on statistics.
+type ConnSuccessRateTracker struct {
+	ConnAttempts *cache.Cache
+	// The success rate of connection attempts, calculated over
+	// ConnAttempts over the interval configured in StartNewConnSuccessRateTracker()
+	ConnSuccessRateStats atomic.Value
+}
+
+// ConnSuccessRateStats represents a timestamped output of computations performed over
+// connection attempts.
```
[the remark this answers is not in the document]
cmoresco-stripe: Done!

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -51,3 +51,55 @@ func NewTestTracker(idle time.Duration) *Tracker {
 
 	return NewTracker(idle, &statsd.NoOpClient{}, logrus.New(), sd, nil)
 }
+
+func TestConnSuccessRateTracker(t *testing.T) {
```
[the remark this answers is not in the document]
cmoresco-stripe: Done!

pkg/smokescreen/conntrack/conn_tracker.go
cmoresco-stripe: To clarify, this parameter isn't the expiry deadline, it's just how quickly stuff gets removed from memory after it expires. It shouldn't have an impact on the calculation, moreso on memory usage. That being said, I think it makes sense to make it a parameter, so if we do want to tweak it we don't need a whole new smokescreen version!

pkg/smokescreen/conntrack/conn_tracker.go
cmoresco-stripe: Sounds good to me! I don't see a great way to detect a misconfiguration before either of these functions are called, since there's no way to tell at startup time whether the caller intends to access these stats. Failing silently instead sgtm!

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +19,111 @@ type Tracker struct {
...
+	CalculatedAt    time.Time
+	ConnSuccessRate float64
+	TotalConns      int
+}
+
+// StartNewConnSuccessRateTracker creates a new ConnSuccessRateTracker with a specific calculation interval at which
+// ConnSuccessRateStats will be recomputed, and a time window to calculate those statistics over.
+// - calculationInterval is how often statistics will be recomputed, in seconds.
+// - calculationWindow is the period that statistics will be calculated over, in seconds.
+// - cleanupInterval is how often expired items (e.g., items older than the calculationWindow) will be deleted from memory, in seconds.
```
ransford-stripe, line 55:
```suggestion
// - calculationInterval is how often statistics will be recomputed.
// - calculationWindow is the period that statistics will be calculated over.
// - cleanupInterval is how often expired items (e.g., items older than the calculationWindow) will be deleted from memory.
```

"in seconds" would be necessary if these values weren't already `time.Duration`, like if they were `int`.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +19,111 @@ type Tracker struct {
...
+		return &stats
+	}
+	return nil
+}
+
+// Removes the port number if it exists. If the string passed in is an IP address, the IP address is returned.
+// If it is a hostname, we return the eTLD + 1 if we are able to parse it, or the unchanged hostname otherwise.
+func normalizeDomainName(requested_host string) string {
+	// Strip port number
+	d := strings.Split(requested_host, ":")[0]
```
ransford-stripe, line 118: You should use `net.SplitHostPort()` here if the string contains a colon.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +19,111 @@ type Tracker struct {
...
+	if net.ParseIP(d) != nil {
+		return d
+	}
+	etldPlusOne, err := publicsuffix.EffectiveTLDPlusOne(d)
```
ransford-stripe, line 122: I love this. TIL!

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -49,5 +49,79 @@ func NewTestTracker(idle time.Duration) *Tracker {
...
+	assert := assert.New(t)
+	tracker := NewTestTracker(time.Second)
+
+	assert.NotPanics(func() { tracker.RecordAttempt("foo.com", true) })
+	stats := tracker.ReportConnectionSuccessRate()
+	assert.Nil(stats)
+
+}
+
+func TestNormalizeDomainName(t *testing.T) {
```
ransford-stripe, line 109: Thanks for adding a test here.

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -49,5 +49,79 @@ func NewTestTracker(idle time.Duration) *Tracker {
...
+
+	var testCases = []struct {
+		domain     string
+		normalized string
+	}{
```
ransford-stripe, line 114: Suggest adding a test case for `"[2001:db8::1]:80"`.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -15,19 +19,111 @@ type Tracker struct {
...
+		SuccessRateTracker: successRateTracker,
+	}
+}
+
+// RecordAttempt stores the result of the most recent connection attempt for a destination.
+func (tr *Tracker) RecordAttempt(dest string, success bool) {
+	if tr.SuccessRateTracker == nil {
+		return
+	}
+	tr.SuccessRateTracker.ConnAttempts.Set(normalizeDomainName(dest), success, cache.DefaultExpiration)
```
ransford-stripe, line 103: This feels like the right place to add hostname normalization, although I wonder if we'll eventually want to be able to toggle normalization (either globally or for a specific `RecordAttempt()`) in case the underlying library is getting it wrong. Anyway, not blocking for now!

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -49,5 +49,79 @@ func NewTestTracker(idle time.Duration) *Tracker {
...
+		waitTime     time.Duration
+		expectedRate float64
+		totalConns   int
+	}{
+		{"fifty-percent-success", []record{{"foo.com", true}, {"bar.com", false}}, 1 * time.Second, 50.0, 2},
+		{"no-records", []record{}, 0 * time.Second, 100.0, 0},
+		// If we wait beyond the calculation window, we should have no records and return the default 100% rate
+		{"expire-records", []record{{"foo.com", true}, {"bar.com", false}}, 3 * time.Second, 100.0, 0},
+		// Only the most recent record for a host is used in the computation
+		{"dedup-hosts", []record{{"foo.com", false}, {"bar.com", false}, {"foo.com", true}}, 1 * time.Second, 50.0, 2},
```
ransford-stripe, line 75: Now that you've added hostname normalization, consider adding a test case that expects `one.foo.com` and `two.foo.com` (and maybe variations with port numbers) to all collapse to `foo.com`.

ransford-stripe (changes requested):
This is looking great. Only small changes suggested after the great changes you already made.

pkg/smokescreen/conntrack/conn_tracker.go
cmoresco-stripe: Me too! One of those magic moments where after a lot of digging you find something that perfectly solves your problem

pkg/smokescreen/conntrack/conn_tracker.go
cmoresco-stripe: of course! thanks :)

pkg/smokescreen/conntrack/conn_tracker.go
cmoresco-stripe: Ack!

ransford-stripe (approved):
Thanks for powering through this.

PR #169, opened 2022-08-22
Report more detail on connection failures

ransford-stripe:
Makes Smokescreen report more detail about connection failures under the `cn.atpt.connect.err` metric, tagged by failure type (`timeout`, `refused`, `reset`, `aborted`).

Tests:
- Adds a  bare-bones `MockMetricsClient` that wraps a `NoOpMetricsClient` with counters so that tests can assert that metrics were or weren't updated.
- Uses the new counting facility to augment existing tests to assert that certain metrics are emitted.
- Adds a test for destination refusing connections.
- Doesn't add tests for ECONNRESET and ECONNABORTED, to avoid adding low-level socket shenanigans to tests and refactoring the guts of all the test harnesses.

A few other small changes:
- Refactors metrics client attachment to `smokescreen.Config`: change into an interface instead of an explicit `*MetricsClient` pointer to allow substitution of a `MockMetricsClient`. Why didn't I simply extend `MetricsClient` to do what I wanted? Because I wanted tests to be able to assert counts of metrics with specific tags, so I needed to store a counter with every metric update, but I didn't want high-cardinality metrics like `cn.atpt.connect.time` (tagged by domain) to eat memory in production for a counting facility we needed only at testing time. Changing to an interface involved hiding `StatsdClient` behind a getter function.
- Adds some descriptive comments.

### DIFF
--- a/pkg/smokescreen/metrics_test.go
+++ b/pkg/smokescreen/metrics_test.go
@@ -62,6 +62,53 @@ func TestMetricsClient(t *testing.T) {
 	})
 }
 
+func TestMockMetricsClient(t *testing.T) {
+	r := require.New(t)
+
+	t.Run("Incr", func(t *testing.T) {
+		m := NewMockMetricsClient()
+		m.Incr("foobar", 1)
+		c, err := m.GetCount("foobar")
+		r.NoError(err)
+		r.Equal(c, uint64(1))
+	})
+
+	t.Run("IncrWithTags", func(t *testing.T) {
+		m := NewMockMetricsClient()
+		tags := []string{"foo", "bar"}
+		m.IncrWithTags("foobar", tags, 1)
+		c, err := m.GetCount("foobar", tags...)
+		r.NoError(err)
+		r.Equal(c, uint64(1))
+	})
+
+	t.Run("IncrWithTags", func(t *testing.T) {
+		m := NewMockMetricsClient()
+		tags := []string{"foo", "bar"}
+		m.IncrWithTags("foobar", tags, 1)
+		c, err := m.GetCount("foobar", tags...)
+		r.NoError(err)
+		r.Equal(c, uint64(1))
+	})
+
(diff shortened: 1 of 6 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -489,10 +490,10 @@ func NormalizeHostPort(hostPort string, forceFQDN bool) (host string, port int,
 // normalized with `normalizeHost` and `normalizePort`.
 //
 // `hostPort` is a bare host or a colon-separated (':') host name and port.
-// If no port is specified, the `scheme`` string is used to find the default
```
cmoresco-stripe, line 492: super minor, but should this and below be a single \` to match the opening \`?

pkg/smokescreen/metrics_test.go
```diff
@@ -56,3 +61,92 @@ func TestMetricsClient(t *testing.T) {
 		r.Error(err)
 	})
 }
+
+// MockMetricsClient is a MetricsClient that counts metric updates.
```
cmoresco-stripe, line 65: I found it surprising that this was defined here, but only used in smokescreen_test.go--have you thought about putting it there instead?

cmoresco-stripe (dismissed):
Looks great, thanks for all the effort you put into making this testable! Comments are optional/suggestions.

pkg/smokescreen/smokescreen.go
ransford-stripe: I blame VS Code! Cleaned up after it in 2e2e15e and c56faa1.

pkg/smokescreen/metrics_test.go
ransford-stripe: As of c4e1bb7, `metrics_test.go` now uses this machinery too. :-)

ransford-stripe:
Thank you for the comments!

r? @cmoresco-stripe

cmoresco-stripe (approved):

PR #170, opened 2022-08-24
Emit metrics from ConnSuccessRateTracker

cmoresco-stripe:
Emits metrics every time the connection success rate is re-computed by the ConnSuccessRateTracker. This is preferable to having the caller of ReportConnectionSuccessRate() do it, as that method may be called many times between re-computations and end up emitting the same metrics repeatedly. 

Emitting these metrics in addition to `cn.atpt.total`, etc. is important, since the ConnSuccessRateTracker performs aggregation and hostname normalization that would be impractical to replicate by postprocessing the raw connection attempt metrics.

This PR does not use the client and testing utilities from https://github.com/stripe/smokescreen/pull/169, as `conntrack` is its own package and to avoid circular imports we would need to restructure the project to move metrics code into its own package as well.

### DIFF
--- a/pkg/smokescreen/conntrack/conn_tracker.go
+++ b/pkg/smokescreen/conntrack/conn_tracker.go
@@ -53,7 +53,7 @@ type ConnSuccessRateStats struct {
 // - calculationInterval is how often statistics will be recomputed.
 // - calculationWindow is the period that statistics will be calculated over.
 // - cleanupInterval is how often expired items (e.g., items older than the calculationWindow) will be deleted from memory.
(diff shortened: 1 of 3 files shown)


ransford-stripe (approved):
This looks functionally OK, but nonblockingly, it's a little strange to have _two_ ersatz metrics client implementations in one codebase. Do you think we should reorganize things a bit? Like put the other package's mock metrics stuff in a package that both the original package and `conntrack` consume?

cmoresco-stripe:
> put the other package's mock metrics stuff in a package that both the original package and `conntrack` consume?

Yeah, that makes sense to me! Conntracker still actually just uses the bare statsd client, so I think we'd actually want to split out all of metrics.go and then migrate conntracker to use our wrapper. Seemed out of scope for what I was trying to do here, but agree it would be nice to have everything using the same client.

PR #171, opened 2022-08-25
Create a Tracker interface so it can be mocked for testing

cmoresco-stripe:
Adds `conntrack.TrackerInterface` that replaces `*contrack.Tracker` in the smokescreen Config. This allows it to be mocked out in unit tests.

### DIFF
--- a/pkg/smokescreen/conntrack/conn_tracker.go
+++ b/pkg/smokescreen/conntrack/conn_tracker.go
@@ -19,7 +19,6 @@ type TrackerInterface interface {
 	MaybeIdleIn(time.Duration) time.Duration
 	NewInstrumentedConn(net.Conn, *logrus.Entry, string, string, string) *InstrumentedConn
 	NewInstrumentedConnWithTimeout(net.Conn, time.Duration, *logrus.Entry, string, string, string) *InstrumentedConn
-
 	Wg() *sync.WaitGroup
 }
 
--- a/pkg/smokescreen/conntrack/conn_tracker.go
+++ b/pkg/smokescreen/conntrack/conn_tracker.go
@@ -168,3 +168,5 @@ func (tr *Tracker) MaybeIdleIn(d time.Duration) time.Duration {
 func (tr *Tracker) Wg() *sync.WaitGroup {
 	return tr.wg
 }
+
+var _ TrackerInterface = &Tracker{}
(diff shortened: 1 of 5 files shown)


pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -13,10 +13,19 @@ import (
 	"golang.org/x/net/publicsuffix"
 )
 
+type TrackerInterface interface {
+	ReportConnectionSuccessRate() *ConnSuccessRateStats
+	RecordAttempt(string, bool)
+	MaybeIdleIn(time.Duration) time.Duration
+	NewInstrumentedConn(net.Conn, *logrus.Entry, string, string, string) *InstrumentedConn
+	NewInstrumentedConnWithTimeout(net.Conn, time.Duration, *logrus.Entry, string, string, string) *InstrumentedConn
+	Wg() *sync.WaitGroup
```
ransford-stripe, line 22: Nonblocking: I wonder if you'd get slightly more intuitive ergonomics by exposing higher-level functions instead of exposing the `WaitGroup` directly. So for example you could have `TrackerInterface.Drain()` or similar and avoid the caller having to think about the mechanism that a tracker uses to implement its connection tracking.

pkg/smokescreen/conntrack/conn_tracker.go
```diff
@@ -155,3 +164,7 @@ func (tr *Tracker) MaybeIdleIn(d time.Duration) time.Duration {
 	})
 	return longest
 }
+
+func (tr *Tracker) Wg() *sync.WaitGroup {
```
ransford-stripe, line 168: While you're thinking about whether `*Tracker` implements `TrackerInterface`, you might want to have the compiler check for you; something like `var _ TrackerInterface = &Tracker{}` at the bottom of this file would do that.

ransford-stripe (approved):
This looks fine as written, but I'm happy to rubber-stamp an interface change if you want to hide the `WaitGroup`.

cmoresco-stripe:
Thanks, I've added the type check!

I agree that the Wg() in the interface is a bit clunky, but I'm not sure refactoring it would be a good use of time since this part of the code is entirely internal to smokescreen and not very heavily used (Wg() only has three non-test callsites!).

ransford-stripe (approved):

PR #172, opened 2022-08-26
Pass destination as a single `HostPort` struct 

lewandowski-stripe:
Additionally move the host+port normalisation to a separate package to make it easier to reuse this functionality in other repositories.


jjiang-stripe (approved):
This is a much cleaner interface! I confirmed that we're pretty much just re-using existing code (with a couple more helpers defined), so there should be no change in behaviour.

jjiang-stripe:
ah there's one failing test because we use `%q` in an error message

jjiang-stripe (changes requested):
i can re-approve once we fix that test!

lewandowski-stripe:
> i can re-approve once we fix that test!

Done. Sorry about the failing tests, I had to rush the PR out of the door before I disappeared for the weekend.

jjiang-stripe (approved):

PR #174, opened 2022-08-29
Move metrics utilities into a package

cmoresco-stripe:
Moves `metrics.go` and `metrics_test.go` into a `metrics` package to allow these utilities to be used by the `conntrack` package. Updates the conntrack package to use this new metrics package.

### DIFF
--- a/pkg/smokescreen/conntrack/conn_tracker_test.go
+++ b/pkg/smokescreen/conntrack/conn_tracker_test.go
@@ -2,12 +2,10 @@ package conntrack
 
 import (
 	"net"
-	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
 
-	"github.com/DataDog/datadog-go/statsd"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stripe/smokescreen/pkg/smokescreen/metrics"
@@ -85,7 +83,7 @@ func TestConnSuccessRateTracker(t *testing.T) {
 
 			sd := atomic.Value{}
 			sd.Store(false)
-			mockMetricsClient := &mockClientInterface{}
+			mockMetricsClient := metrics.NewMockMetricsClient()
 			tracker := NewTracker(time.Second, metrics.NewNoOpMetricsClient(), logrus.New(), sd, StartNewConnSuccessRateTracker(500*time.Millisecond, 2*time.Second, 10*time.Second, mockMetricsClient))
 
 			for _, record := range tc.additions {
@@ -98,8 +96,13 @@ func TestConnSuccessRateTracker(t *testing.T) {
 			assert.InDelta(tc.expectedRate, stats.ConnSuccessRate, 0.01)
 			assert.Equal(tc.totalConns, stats.TotalConns)
 
-			assert.Equal(tc.expectedRate, mockMetricsClient.getValue("cn.atpt.distinct_domains_success_rate"))
-			assert.Equal(float64(tc.totalConns), mockMetricsClient.getValue("cn.atpt.distinct_domains"))
+			v, err := mockMetricsClient.GetValues("cn.atpt.distinct_domains_success_rate")
+			assert.NoError(err)
+			assert.Equal(tc.expectedRate, v[len(v)-1])
+
+			v, err = mockMetricsClient.GetValues("cn.atpt.distinct_domains")
+			assert.NoError(err)
+			assert.Equal(tc.totalConns, int(v[len(v)-1]))
 
 		})
 	}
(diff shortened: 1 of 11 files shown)


pkg/smokescreen/metrics/mocks.go
```diff
@@ -174,6 +75,31 @@ func (m *MockMetricsClient) GetCount(metric string, tags ...string) (uint64, err
 	return i, nil
 }
 
+// GetValues returns the values stored for a metric metric has been updated since the MockMetricsClient was
+// created. To support GetValues being called with or without tags for a given metric, tagged metrics
+// are counted twice: once for the untagged metric ("foo") and once for the metric with its tags
+// sorted("foo [a b c]").
```
ransford-stripe, line 81:
```suggestion
// GetValues returns the values stored for a metric that has been updated since the MockMetricsClient was
// created.
```
(The bit about double-counting seemed not quite right here.)

pkg/smokescreen/metrics/metrics.go
```diff
@@ -56,6 +56,9 @@ type MetricsClientInterface interface {
 	AddMetricTags(string, []string) error
 	Incr(string, float64) error
 	IncrWithTags(string, []string, float64) error
+	Gauge(string, float64, float64) error
```
ransford-stripe, line 59: No `GaugeWithTags`? Just wondering for consistency's sake.

pkg/smokescreen/metrics/metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+func TestMetricsTags(t *testing.T) {
+	r := require.New(t)
+
+	t.Run("add custom tags", func(t *testing.T) {
+		metric := "acl.allow"
+		mc := NewNoOpMetricsClient()
+
+		err := mc.AddMetricTags(metric, []string{"globalize"})
```
ransford-stripe, line 17: Nonblocking note: it'd be nice to have a better comment on `AddMetricTags` to indicate that (AFAICT) tags added in that way will _always_ be emitted along with the metrics they're attached to.

pkg/smokescreen/metrics/metrics_test.go
```diff
@@ -0,0 +1,160 @@
...
+		r.Equal(tags[0], "globalize")
+
+		err = mc.AddMetricTags(metric, []string{"ignore"})
+		r.NoError(err)
+
+		tags = mc.GetMetricTags(metric)
+		r.Len(tags, 2)
+	})
+
+	t.Run("add invalid tags", func(t *testing.T) {
```
ransford-stripe, line 31:
```suggestion
	t.Run("add tags to a nonexistent metric", func(t *testing.T) {
```

pkg/smokescreen/conntrack/conn_tracker_test.go
```diff
@@ -84,8 +83,8 @@ func TestConnSuccessRateTracker(t *testing.T) {
 
 			sd := atomic.Value{}
 			sd.Store(false)
-			mockMetricsClient := &mockClientInterface{}
-			tracker := NewTracker(time.Second, &statsd.NoOpClient{}, logrus.New(), sd, StartNewConnSuccessRateTracker(500*time.Millisecond, 2*time.Second, 10*time.Second, mockMetricsClient))
+			mockMetricsClient := metrics.NewMockMetricsClient()
+			tracker := NewTracker(time.Second, metrics.NewNoOpMetricsClient(), logrus.New(), sd, StartNewConnSuccessRateTracker(500*time.Millisecond, 2*time.Second, 10*time.Second, mockMetricsClient))
```
ransford-stripe, line 87: Nonblocking: I realize this is just a translation of the previous code, but this line is getting a bit unwieldy and it's a bit confusing to call `metrics.NewNoOpMetricsClient()` right after you create a `mockMetricsClient`. This line of code seems to use two metrics clients, and it's not obvious that that's intentional.

ransford-stripe (approved):
Only a few nonblocking suggestions. Thanks for doing this refactoring.

pkg/smokescreen/metrics/mocks.go
cmoresco-stripe: Well, the "foo"/"foo [a b c" bit is still relevant, since this function does record values for both--I can update the language to get rid of the word "counting" though

pkg/smokescreen/metrics/metrics_test.go
cmoresco-stripe: Updated the comment!

pkg/smokescreen/conntrack/conn_tracker_test.go
cmoresco-stripe: Understand the confusion! I made the parameters multi-line and added a comment to explain why we're using a no-op here.

pkg/smokescreen/metrics/metrics.go
cmoresco-stripe: It feels a little weird to me to add dead code just for the sake of having it there, especially since we're already covering such a limited set of the statsd client interface with this API (and because these methods are pretty trivial to implement, I don't think it'll get it in the way of anyone who wants to add tags to their gauge down the line)

cmoresco-stripe:
Thanks for the thoughtful feedback!

ransford-stripe (approved):
Great updates!

PR #177, opened 2022-08-30
Permit underscores in requested domain names

lewandowski-stripe:
The practice shows that the use of underscores in domain names is unfortunately way too common to be ignored.


jjiang-stripe (approved):
confirmed that the options we're setting seem to match the default, except we disable STD3 rules which are currently blocking underscores.

PR #178, opened 2022-09-20
Normalize domains in ACL validation

jjiang-stripe:
cc @stripe/smokescreen-reviewers 

This PR updates the ACL validation to ensure folks use the normalized form in their ACLs. 

This has a couple benefits:
* Currently we normalize the request but don't normalize the ACL entry, so if you specified `éxämple.com` in your ACL, it would not actually match a request to `éxämple.com`. By ensuring all ACL entries are normalized, we only check equality on normalized hostnames.
* This helps prevent any ambiguity in the ACL; `аррӏе.com` is actually all Cyrillic letters that look like Latin letters. This PR enforces that it'd be written as `xn--80ak6aa92e.com` instead.


lewandowski-stripe (approved):
LGTM!

PR #180, opened 2022-11-22
Add VerifyRequestHandler config option

param-stripe:
This PR adds support for passing in a callback method to verify requests. We're doing this because we want to verify requests internally at Stripe.

Motivation: https://jira.corp.stripe.com/browse/SECURE_FRAMEWORKS-2919

Test plan: I've added tests for both HTTP and HTTPS cases.

r? @cds2-stripe

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -79,6 +79,10 @@ type Config struct {
 	// ranges by default (exempting loopback and unicast ranges)
 	// This setting can be used to configure Smokescreen with a blocklist, rather than an allowlist
 	UnsafeAllowPrivateRanges bool
+
+	// Customer handler for verifying requests, users can pass in custom methods to verify requests based
+	// on headers etc.
+	VerifyRequestHandler func(*http.Request) error
 }
 
 type missingRoleError struct {
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -82,6 +82,7 @@ type Config struct {
 
 	// Custom handler for users to allow running code per requests, users can pass in custom methods to verify requests based
 	// on headers, code for metrics etc.
+	// If the handler returns an error, smokescreen will deny the request.
 	CustomRequestHandler func(*http.Request) error
 }
 
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -80,9 +80,9 @@ type Config struct {
 	// This setting can be used to configure Smokescreen with a blocklist, rather than an allowlist
 	UnsafeAllowPrivateRanges bool
 
-	// Customer handler for verifying requests, users can pass in custom methods to verify requests based
-	// on headers etc.
-	VerifyRequestHandler func(*http.Request) error
+	// Custom handler for users to allow running code per requests, users can pass in custom methods to verify requests based
+	// on headers, code for metrics etc.
+	CustomRequestHandler func(*http.Request) error
 }
 
 type missingRoleError struct {
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -475,9 +475,9 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		sctx.logger.WithField("url", req.RequestURI).Debug("received HTTP proxy request")
 
-		// verify the request based on the handler, if it exists
-		if config.VerifyRequestHandler != nil {
-			err = config.VerifyRequestHandler(pctx.Req)
+		// Call the custom request handler if it exists
+		if config.CustomRequestHandler != nil {
+			err = config.CustomRequestHandler(pctx.Req)
 			if err != nil {
 				pctx.Error = denyError{err}
 				return req, rejectResponse(pctx, pctx.Error)
@@ -619,9 +619,9 @@ func handleConnect(config *Config, pctx *goproxy.ProxyCtx) (string, error) {
 		return "", pctx.Error
 	}
 
-	// verify the request based on the handler, if it exists
-	if config.VerifyRequestHandler != nil {
-		err = config.VerifyRequestHandler(pctx.Req)
+	// Call the custom request handler if it exists
+	if config.CustomRequestHandler != nil {
+		err = config.CustomRequestHandler(pctx.Req)
 		if err != nil {
 			pctx.Error = denyError{err}
 			return "", pctx.Error
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -475,6 +475,15 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		sctx.logger.WithField("url", req.RequestURI).Debug("received HTTP proxy request")
 
+		// verify the request based on the handler, if it exists
+		if config.VerifyRequestHandler != nil {
+			err = config.VerifyRequestHandler(pctx.Req)
+			if err != nil {
+				pctx.Error = denyError{err}
+				return req, rejectResponse(pctx, pctx.Error)
+			}
+		}
+
 		sctx.decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, req, destination)
 
 		// Returning any kind of response in this handler is goproxy's way of short circuiting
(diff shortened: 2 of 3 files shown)


pkg/smokescreen/smokescreen_test.go
```diff
@@ -1019,6 +1021,121 @@ func TestRejectResponseHandler(t *testing.T) {
...
+				r.Nil(resp)
+				r.Contains(err.Error(), "Request rejected by proxy")
+			} else {
+				r.NoError(err)
+				r.Equal(200, resp.StatusCode)
+			}
+		}
+	})
+
+	t.Run("test that it works for HTTP (invalid)", func(t *testing.T) {
```
sergeyrud-stripe, line 1086: What does `(invalid)` mean? Since we're testing both valid and invalid use cases here, I'd image we don't need it here.
```suggestion
	t.Run("test that it works for HTTP", func(t *testing.T) {
```

pkg/smokescreen/smokescreen_test.go
```diff
@@ -1019,6 +1021,121 @@ func TestRejectResponseHandler(t *testing.T) {
...
+		proxySrv := proxyServer(cfg)
+		r.NoError(err)
+		defer proxySrv.Close()
+
+		// Create a http.Client that uses our proxy
+		client, err := proxyClient(proxySrv.URL, nil)
+		r.NoError(err)
+
+		for _, testCase := range testCases {
+			// Send a request that should be blocked
```
sergeyrud-stripe, line 1121: Is this comment up-to-date?
```suggestion
```

pkg/smokescreen/smokescreen_test.go
```diff
@@ -1069,7 +1186,7 @@ func proxyServer(conf *Config) *httptest.Server {
 	return httptest.NewServer(proxy)
 }
 
-func proxyClient(proxy string) (*http.Client, error) {
+func proxyClient(proxy string, h http.Header) (*http.Client, error) {
```
sergeyrud-stripe, line 1189: (Optional suggestion) `h` or `header` might be too general. Maybe we can call it `proxyHeader` or `proxyConnectHeader`?
```suggestion
func proxyClient(proxy string, proxyConnectHeader http.Header) (*http.Client, error) {
```

On the other hand, considering that we use this header only in one place, can/should we leave this func as-is and update the client instead of passing `nil` everywhere?
```go
client, err := proxyClient(proxy.URL)
// check the err
client.Transport.ProxyConnectHeader = h
```

Just FYI, for our internal use case we're going to pass 2 extra headers instead of one extra header. So if we were to rename the variable, maybe it would be better to underscore that it could be plural.

pkg/smokescreen/smokescreen_test.go
```diff
@@ -1019,6 +1021,121 @@ func TestRejectResponseHandler(t *testing.T) {
...
+
+		// Create a http.Client that uses our proxy
+		client, err := proxyClient(proxySrv.URL, nil)
+		r.NoError(err)
+
+		for _, testCase := range testCases {
+			// Send a request that should be blocked
+			req, err := http.NewRequest("GET", remote.URL, nil)
+			r.NoError(err)
+			req.Header.Set(testHeader, testCase.header)
```
sergeyrud-stripe, line 1124: We're passing an extra header to conduct some verification here. We should remove it from the request before proxying it further as we don't want to leak internal details or pass unnecessary information.

Could you update the test to demonstrate that it's possible to remove the header from the original request before proxying it?

pkg/smokescreen/smokescreen.go
```diff
@@ -475,6 +475,15 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		sctx.logger.WithField("url", req.RequestURI).Debug("received HTTP proxy request")
 
+		// verify the request based on the handler, if it exists
+		if config.VerifyRequestHandler != nil {
+			err = config.VerifyRequestHandler(pctx.Req)
```
sergeyrud-stripe, line 480: I noticed that we for cleartext requests in this handler we use `req` instead of `pctx.Req`. What's the difference between these two variables? Should we pass `req` here instead for consistency? You can see that above we mutate `req`
```go
req.Header.Del(roleHeader)
req.Header.Del(traceHeader)
```
We'd want to do the same in our handler. I.e. remove extra headers.

sergeyrud-stripe (changes requested):

pkg/smokescreen/config.go
```diff
@@ -79,6 +79,10 @@ type Config struct {
 	// ranges by default (exempting loopback and unicast ranges)
 	// This setting can be used to configure Smokescreen with a blocklist, rather than an allowlist
 	UnsafeAllowPrivateRanges bool
+
+	// Customer handler for verifying requests, users can pass in custom methods to verify requests based
+	// on headers etc.
+	VerifyRequestHandler func(*http.Request) error
```
sergeyrud-stripe, line 85: (optional suggestion) I wonder if we want to make it general (eg. `CustomRequestHandler` or `RequestHook`) so it can be used for different use cases. I.e. some folks may use it for verification, others may pass additional data for metrics, etc.
```suggestion
	CustomRequestHandler func(*http.Request) error
```

pkg/smokescreen/smokescreen_test.go
param-stripe: oh whoops, yeah, this is a typo from me changing the tests.

pkg/smokescreen/smokescreen_test.go
param-stripe: The types are a bit problematic here. `client.Transport` is typed as `RoundTripper`, not `http.Transport`, so what you suggest leads to a compile error (unless I'm misunderstanding something). I'll rename it to `proxyConnectHeaders` for clarity.

pkg/smokescreen/smokescreen.go
param-stripe: cc @cds2-stripe in case you have insight.

pkg/smokescreen/smokescreen_test.go
param-stripe: Removing the header in the handler now.

param-stripe: https://github.com/stripe/smokescreen/pull/180/commits/382d64a93b264ff460417837bf4fe4efe130a281

sergeyrud-stripe (approved):

pkg/smokescreen/config.go
```diff
@@ -79,6 +79,10 @@ type Config struct {
 	// ranges by default (exempting loopback and unicast ranges)
 	// This setting can be used to configure Smokescreen with a blocklist, rather than an allowlist
 	UnsafeAllowPrivateRanges bool
+
+	// Custom handler for users to allow running code per requests, users can pass in custom methods to verify requests based
+	// on headers, code for metrics etc.
+	CustomRequestHandler func(*http.Request) error
```
cds2-stripe, line 85: Could you include what happens when an error is returned in this comment?

pkg/smokescreen/smokescreen.go
```diff
@@ -475,6 +475,15 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 
 		sctx.logger.WithField("url", req.RequestURI).Debug("received HTTP proxy request")
 
+		// Call the custom request handler if it exists
+		if config.CustomRequestHandler != nil {
+			err = config.CustomRequestHandler(req)
+			if err != nil {
+				pctx.Error = denyError{err}
```
cds2-stripe, line 482: WDYT about creating a new error type here?

pkg/smokescreen/smokescreen_test.go
```diff
@@ -1069,7 +1198,7 @@ func proxyServer(conf *Config) *httptest.Server {
 	return httptest.NewServer(proxy)
 }
 
-func proxyClient(proxy string) (*http.Client, error) {
+func proxyClient(proxy string, proxyConnectHeaders http.Header) (*http.Client, error) {
```
cds2-stripe, line 1201: nit: Instead of reusing this function and having to update all of the existing call sites to pass in a nil `proxyConnectHeaders` my preference would be to create a new `proxyClientWithConnectHeaders` function that wraps the original `proxyClient` function.

cds2-stripe (changes requested):

pkg/smokescreen/smokescreen.go
param-stripe: I don't feel too strongly about this, but I was pretty intuitively happy with denyErrors containing the error returned by the custom handler. I'm ok with keeping as is, unless you have stronger opinions.

pkg/smokescreen/smokescreen_test.go
param-stripe: done!

pkg/smokescreen/config.go
param-stripe: for sure, done!

param-stripe:
ptal @cds2-stripe

cds2-stripe (approved):

PR #183, opened 2022-12-30
Log resolved IP address

natashadada:
Adds resolved ip address to logs.


natashadada:
It looks like this is already being logged as `outbound_remote_addr`

PR #184, opened 2023-01-05
Add a header for metadata to log

natashadada:
Adds a header "X-Smokescreen-Log-Metadata" that works similarly to the  "X-Smokescreen-Trace-ID" header. The log metadata header takes in a string that is included in proxy logs under "log_metadata"

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -59,6 +59,7 @@ const (
 	CanonicalProxyDecision   = "CANONICAL-PROXY-DECISION"
 	LogFieldConnEstablishMS  = "conn_establish_time_ms"
 	LogFieldDNSLookupTime    = "dns_lookup_time_ms"
+	LogFieldLogMetadata      = "log_metadata"
 )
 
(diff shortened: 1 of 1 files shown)


natashadada:
@ransford-stripe I saw you've been maintaining this repo recently. Would you be willing to review this change?

jjiang-stripe:
Hi @natashadada, thanks for submitting this PR! Could you explain what you're expecting to log with `X-Smokescreen-Log-Metadata`? 

I'm wondering if it makes more sense to emit logs on the client, and use `X-Smokescreen-Trace-ID` to correlate the logs between your client and Smokescreen.

natashadada:
> Hi @natashadada, thanks for submitting this PR! Could you explain what you're expecting to log with `X-Smokescreen-Log-Metadata`?
> 
> I'm wondering if it makes more sense to emit logs on the client, and use `X-Smokescreen-Trace-ID` to correlate the logs between your client and Smokescreen.

Sorry for the very late reply here. Correlating trace id will work for us, thank you!

PR #188, opened 2023-04-21
Tag acl. metrics with requested_host

kevinv-stripe:
This change tags `acl.` metrics (`acl.deny`, `acl.allow`, ...) with the `requested_host`. The tag can be used to ignore certain known denials in alerting or count the number of requests to a given host.


kevinv-stripe:
Closing since this would greatly increase cardinality of metrics which, depending on the size of the smokescreen deployment, would impact observability platforms.

PR #189, opened 2023-05-24
Move the custom request handler call after the main ACL check

sergeyrud-stripe:
Running a custom request handler after a successful ACL check is more useful for certain use cases. I.e. if a request is denied by `checkIfRequestShouldBeProxied`, this request might be irrelevant for `CustomRequestHandler`.


kevinv-stripe (approved):
LGTM. Thanks for the PR! 

For other smokescreen users: if this change breaks your implementation please file an issue and we'll address it ASAP.

PR #192, opened 2023-06-23
Use local server instead of httpbin in test

cmoresco-stripe:
The `TestClearsErrorHeader` test has appeared flaky because httpbin.org has been timing out. This PR replaces the dependency on httpbin.org with a local golang httptest.Server that sets the desired headers.


jjiang-stripe (approved):
oooh nice catch!!

PR #196, opened 2023-07-26
add AcceptResponseHandler to modify accepted responses

cmoresco-stripe:
This change adds an AcceptReponseHandler to the config, which provides a hook to modify an accepted CONNECT response before it is returned to the client. This complements the existing RejectResponseHandler.

### DIFF
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -82,6 +82,9 @@ type Config struct {
 	// Customer handler to allow clients to modify reject responses
 	RejectResponseHandler func(*http.Response)
 
+	// Customer handler to allow clients to modify accept responses
+	AcceptResponseHandler func(*http.Response)
+
 	// UnsafeAllowPrivateRanges inverts the default behavior, telling smokescreen to allow private IP
 	// ranges by default (exempting loopback and unicast ranges)
 	// This setting can be used to configure Smokescreen with a blocklist, rather than an allowlist
--- a/pkg/smokescreen/config.go
+++ b/pkg/smokescreen/config.go
@@ -79,10 +79,10 @@ type Config struct {
 	// Custom Dial Timeout function to be called
 	ProxyDialTimeout func(ctx context.Context, network, address string, timeout time.Duration) (net.Conn, error)
 
-	// Customer handler to allow clients to modify reject responses
+	// Custom handler to allow clients to modify reject responses
 	RejectResponseHandler func(*http.Response)
 
-	// Customer handler to allow clients to modify accept responses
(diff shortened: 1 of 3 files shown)


pkg/smokescreen/config.go
```diff
@@ -82,6 +82,9 @@ type Config struct {
 	// Customer handler to allow clients to modify reject responses
 	RejectResponseHandler func(*http.Response)
 
+	// Customer handler to allow clients to modify accept responses
```
jjiang-stripe, line 85: Reading this I'm wondering if the comment was supposed to say "custom" 🤔

pkg/smokescreen/smokescreen.go
```diff
@@ -541,10 +541,13 @@ func BuildProxy(config *Config) *goproxy.ProxyHttpServer {
 	proxy.OnResponse().DoFunc(func(resp *http.Response, pctx *goproxy.ProxyCtx) *http.Response {
 		sctx := pctx.UserData.(*smokescreenContext)
 
-		if resp != nil && resp.Header.Get(errorHeader) != "" {
-			if pctx.Error == nil && sctx.decision.allow {
+		if resp != nil && pctx.Error == nil && sctx.decision.allow {
```
jjiang-stripe, line 544: I think we actually want to keep these conditionals the same, or hoist up the error header check outside of this `if`. 

We're trying to delete any response headers that we didn't set (e.g. the upstream server could set a fake `X-Smokescreen-Error` header), but I think we'll miss that deletion with this change.

jjiang-stripe (changes requested):

pkg/smokescreen/smokescreen.go
cmoresco-stripe: I'm not sure I see how the deletion logic changes? We're replacing 

```
if a && b {
  if c && d {
    <delete>
  }
}
```

with

```
if a && c && d {
  if b {
    <delete>
  }
}
```

Either way, all four conditionals have to be true for the header to be deleted

pkg/smokescreen/config.go
cmoresco-stripe: Probably!

pkg/smokescreen/smokescreen.go
jjiang-stripe: oh whoops you're right! it's too early 🥲

jjiang-stripe (approved):

pkg/smokescreen/smokescreen.go
cmoresco-stripe: :awake-but-at-what-cost:

PR #199, opened 2023-08-02
Use AcceptResponseHandler in goproxy https CONNECT hook

cmoresco-stripe:
Wire up the AcceptResponseHandler to the new goproxy hook that was added in https://github.com/stripe/goproxy/pull/21

This does require changing the type of the AcceptResponseHandler to include a context field, but since this handler was only added to the config ~days ago this shouldn't be disruptive.

Smokescreen takes a function that uses a smokescreenContext and wraps it in a function that takes a goproxy.ProxyCtx before passing it to the underlying goproxy, so that we don't leak any goproxy interfaces in the smokescreen API (remain eternally hopeful that we'll strip out goproxy entirely someday...)

Writing a unit test would require the new go 1.20 [http.OnProxyConnectResponse](https://github.com/golang/go/issues/54299) to hook into the proxy connect response, and I don't think we're ready to upgrade smokescreen to go 1.20 yet, so the https logic here is not covered by unit tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/rs/xid v1.2.1
 	github.com/sirupsen/logrus v1.9.0
 	github.com/stretchr/testify v1.8.0
-	github.com/stripe/goproxy v0.0.0-20220308202309-3f1dfba6d1a4
+	github.com/stripe/goproxy v0.0.0-20230801191332-fabc3ecb7251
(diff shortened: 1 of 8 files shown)


jjiang-stripe (approved):
ah sorry I thought I hit approve already :')

this seems reasonable to me! we're just hooking into the `ConnectRespHandler` functionality we added.

PR #200, opened 2023-08-04
Export SmokescreenContext type

cmoresco-stripe:
In https://github.com/stripe/smokescreen/pull/199 we added a hook that will be called on a CONNECT response and takes a smokescreenContext. In order for callers to be able to write functions that use smokescreenContext, we need to export the type.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -250,7 +250,7 @@ func dialContext(ctx context.Context, network, addr string) (net.Conn, error) {
 	if !ok {
 		return nil, fmt.Errorf("dialContext missing required *SmokescreenContext")
 	}
-	d := sctx.decision
+	d := sctx.Decision
 
 	// If an address hasn't been resolved, does not match the original outboundHost,
 	// or is not tcp we must re-resolve it before establishing the connection.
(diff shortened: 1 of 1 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -70,7 +70,7 @@ type aclDecision struct {
 	enforceWouldDeny                    bool
 }
 
-type smokescreenContext struct {
+type SmokescreenContext struct {
 	cfg           *Config
 	start         time.Time
 	decision      *aclDecision
```
jjiang-stripe, line 76: Do we also need to export this field?

jjiang-stripe (changes requested):

pkg/smokescreen/smokescreen.go
cmoresco-stripe: Good call! Also ResolvedAddr. Thanks!

pkg/smokescreen/smokescreen.go
```diff
@@ -63,17 +63,17 @@ const (
...
 	allow                               bool
 	enforceWouldDeny                    bool
 }
 
-type smokescreenContext struct {
+type SmokescreenContext struct {
 	cfg           *Config
 	start         time.Time
-	decision      *aclDecision
+	decision      *ACLDecision
```
jjiang-stripe, line 76:
```suggestion
	Decision      *ACLDecision
```

jjiang-stripe (changes requested):
final lil capitalization bit

jjiang-stripe (approved):

PR #201, opened 2023-08-04
Fix outbound IP logging

jjiang-stripe:
I was staring at this chunk of code earlier trying to figure out how the outbound remote IP was logged, and realized there was actually a little coding error here. `fields` in `pkg/smokescreen/smokescreen.go` was in a closure, so it wasn't writing the fields, and now that it is, we can remove a bit of duplicate code in `pkg/smokescreen/conntrack/instrumented_conn.go`.

I deployed this change to our QA environment internally and verified that we still see `outbound_remote_addr` on connection close, as well as `outbound_local_addr`

### DIFF
--- a/pkg/smokescreen/conntrack/instrumented_conn.go
+++ b/pkg/smokescreen/conntrack/instrumented_conn.go
@@ -17,7 +17,6 @@ const (
 	LogFieldDuration        = "duration"
 	LogFieldError           = "error"
 	LogFieldLastActivity    = "last_activity"
-	LogFieldOutboundAddr    = "outbound_remote_addr"
 	CanonicalProxyConnClose = "CANONICAL-PROXY-CN-CLOSE"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/smokescreen/conntrack/instrumented_conn.go
```diff
@@ -114,19 +113,13 @@ func (ic *InstrumentedConn) Close() error {
...
-	}
-
 	ic.logger.WithFields(logrus.Fields{
 		LogFieldBytesIn:      ic.BytesIn,
 		LogFieldBytesOut:     ic.BytesOut,
 		LogFieldEndTime:      end.UTC(),
 		LogFieldDuration:     duration,
 		LogFieldError:        errorMessage,
 		LogFieldLastActivity: time.Unix(0, atomic.LoadInt64(ic.LastActivity)).UTC(),
-		LogFieldOutboundAddr: outboundAddr,
```
jjiang-stripe, line 129: this is removed because it's already on the logger as `LogFieldOutRemoteAddr`

cds2-stripe (approved):

PR #203, opened 2023-08-21
Allow listen address specification for prometheus metrics

ne-bknn:
Hey! 

Stumbled upon a problem that smokescreen will expose prometheus metrics for all of the interfaces; there is no way to change listen address, so the only way to limit metrics access is by using iptables, which is not desireable in my situation.

Hence, this patch.


cds2-stripe (approved):
Sorry for the delay in reviewing this - this looks great! Appreciate the patch.

PR #207, opened 2023-11-08
bump minimum Go version to 1.20

cds2-stripe:
Per our docs, we only support the most recent two versions of Go. This is currently 1.20 and 1.21.

r? @jjiang-stripe


jjiang-stripe (approved):
🙏

PR #208, opened 2023-11-14
Configurable http and https proxy addrs

xieyuxi-stripe:
### Summary
Update the stripe/goproxy dependency to include https://github.com/stripe/goproxy/pull/22. Allow setting the httpProxyAddr and httpsProxyAddr directly in smokescreen's configuration. When the TransportHttpProxyAddr and TransportHttpsProxyAddr in the smokescreen configuration is non-empty, use these proxy address to create the `NewProxyHttpServer` for smokescreen. If these addresses are not set, use the default values from environment. 

### Motivation
stripe/smokescreen uses this library for the proxy server. Currently, the only way to set the transport proxy for smokescreen are [setting the proxy from environment](https://github.com/stripe/goproxy/blob/master/proxy.go#L192), which requires us to set the environment variables like HTTP_PROXY, HTTPS_PROXY. We would like to have a more robust way of configuring the smokescreen server's transport proxy. Thus we are adding the transport proxy address as another configurations for the smokescreen server and the goproxy library it uses underneath.

### Test plan
- [x] New and existing unit tests
- [x] Vendored the smokescreen change in stripe-internal repo and tested the services that depends on it in QA

### DIFF
--- a/cmd/smokescreen.go
+++ b/cmd/smokescreen.go
@@ -94,11 +94,11 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
 			Value: "/metrics",
 			Usage: "Expose prometheus metrics on `ENDPOINT`. Requires --expose-prometheus-metrics to be set. Defaults to \"/metrics\"",
 		},
-    cli.StringFlag{
-      Name: "prometheus-listen-ip",
-      Value: "0.0.0.0",
-      Usage: "Listen for prometheus metrics on interface with address IP. Requires --expose-prometheus-metrics to be set. Defaults to \"0.0.0.0\"",
-    },
+		cli.StringFlag{
+			Name: "prometheus-listen-ip",
+			Value: "0.0.0.0",
+			Usage: "Listen for prometheus metrics on interface with address IP. Requires --expose-prometheus-metrics to be set. Defaults to \"0.0.0.0\"",
+		},
 		cli.StringFlag{
 			Name:  "prometheus-port",
 			Value: "9810",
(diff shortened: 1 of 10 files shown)


cmd/smokescreen.go
```diff
@@ -146,6 +146,16 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
 			Name:  "unsafe-allow-private-ranges",
 			Usage: "Allow private ip ranges by default",
 		},
+		cli.StringFlag{
+			Name:  "transport-http-proxy-addr",
+			Value: "",
+			Usage: "Set the smokescreen's HTTP transport proxy address",
```
cds2-stripe, line 152: Could we change this to:
```
Set Smokescreen's upstream HTTP proxy address
```

cmd/smokescreen.go
```diff
@@ -146,6 +146,16 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
...
+		},
+		cli.StringFlag{
+			Name:  "transport-https-proxy-addr",
+			Value: "",
+			Usage: "Set the smokescreen's HTTPS transport proxy address",
```
cds2-stripe, line 157: Same as above but for HTTPS

cmd/smokescreen.go
```diff
@@ -146,6 +146,16 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
 			Name:  "unsafe-allow-private-ranges",
 			Usage: "Allow private ip ranges by default",
 		},
+		cli.StringFlag{
+			Name:  "transport-http-proxy-addr",
```
cds2-stripe, line 150: We should change this to: `upstream-http-proxy-addr`

cmd/smokescreen.go
```diff
@@ -146,6 +146,16 @@ func NewConfiguration(args []string, logger *log.Logger) (*smokescreen.Config, e
...
+			Value: "",
+			Usage: "Set the smokescreen's HTTP transport proxy address",
+		},
+		cli.StringFlag{
+			Name:  "transport-https-proxy-addr",
```
cds2-stripe, line 155: `upstream-https-proxy-addr`

pkg/smokescreen/config.go
```diff
@@ -73,6 +73,10 @@ type Config struct {
 	TransportMaxIdleConns        int
 	TransportMaxIdleConnsPerHost int
 
+	// There are the http and https address for the transport proxy
```
cds2-stripe, line 76:
```
UpstreamHttpProxyAddr  string
UpstreamHttpsProxyAddr string
```

cds2-stripe (changes requested):
Just a couple of small variable changes!

cds2-stripe (approved):

PR #210, opened 2023-12-01
Update docs to disambiguate ACL vs --deny-address behavior

cmoresco-stripe:
Update README language to clarify that the ACL is only applied to hostnames pre-DNS resolution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,6 @@ Three policies are supported:
 
 A host can be specified with or without a globbing prefix. The host (without the globbing prefix) must be in Punycode to prevent ambiguity.
 
-
 | host                | valid   |
 | ------------------- | ------- |
 | `example.com`       | yes     |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,10 +117,13 @@ func main() {
...
 ```
+### IP Filtering
 
-### ACLs
+To control the routing of requests to specific IP addresses or IP blocks, use the `deny-address`, `allow-address`, `deny-range`, and `allow-range` options in the config. 
 
-An ACL can be described in a YAML formatted file. The ACL, at its top-level, contains a list of services as well as a default behavior.
+### Hostname ACLs
+
+An Hostname ACL can be described in a YAML formatted file. The ACL, at its top-level, contains a list of services as well as a default behavior.
```
alecb-stripe, line 126: nit: "A hostname ACL"

alecb-stripe (approved):
LG, soft suggestion to also modify the paragraph in the intro block:

https://github.com/stripe/smokescreen/blob/6d836e26b7b273ada8c891332e9f15d7ad3aabaa/README.md?plain=1#L6

since imo as-written it makes it sound like smokescreen's primary functionality is IP-based filtering.

Which I think maybe contributes to people thinking even the hostname-based filtering is actually IP filtering or something (i.e. they think we resolve the hostnames and block access to those IPs, or something?)

PR #213, opened 2024-02-12
Add support for Smokescreen -> HTTPS CONNECT Proxy ACLs

pspieker-stripe:
As part of go/dist-egress, we want to be able to talk to an external-to-Stripe HTTPS CONNECT proxy. We'd also like to default block traffic going to any external-to-Stripe HTTPS CONNECT proxy, unless it's on a given allow list (basically a CONNECT proxy URL ACL). 

This PR adds support for such an allow list in smokescreen. 

Testing via:
```
cd pkg/smokescreen
go test -v -run ^TestCONNECTProxyACLs$
dlv test -- -test.run=^TestCONNECTProxyACLs$
```

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -934,10 +934,9 @@ func checkACLsForRequest(config *Config, req *http.Request, destination hostport
 	// a _subsequent_ proxy to use for the CONNECT request. This is used to allow traffic
 	// flow as in: client -(TLS)-> smokescreen -(TLS)-> external proxy -(TLS)-> destination.
 	// Without this header, there's no way for the client to specify a subsequent proxy.
-	var connectProxyHost string
-	if connectProxyHostSlice := req.Header.Get("X-Upstream-Https-Proxy"); len(connectProxyHostSlice) > 0 {
-		connectProxyHost = string(connectProxyHostSlice[0])
-	}
+	// Also note - Get returns the first value for a given header, or the empty string,
+	// which is the behavior we want here.
+	connectProxyHost := req.Header.Get("X-Upstream-Https-Proxy")
 
 	ACLDecision, err := config.EgressACL.Decide(role, destination.Host, connectProxyHost)
 	decision.project = ACLDecision.Project
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -639,11 +639,9 @@ func handleConnect(config *Config, pctx *goproxy.ProxyCtx) (string, error) {
 		return "", pctx.Error
 	}
 
-	/*
-		checkIfRequestShouldBeProxied can return an error if either the resolved address is disallowed,
-		or if there is a DNS resolution failure, or if the subsequent proxy host (specified by the
-		X-Https-Upstream-Proxy header in the CONNECT request to _this_ proxy) is disallowed.
-	*/
+	// checkIfRequestShouldBeProxied can return an error if either the resolved address is disallowed,
+	// or if there is a DNS resolution failure, or if the subsequent proxy host (specified by the
+	// X-Https-Upstream-Proxy header in the CONNECT request to _this_ proxy) is disallowed.
 	sctx.Decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, pctx.Req, destination)
 	if pctx.Error != nil {
 		// DNS resolution failure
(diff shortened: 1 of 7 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -932,7 +935,18 @@ func checkACLsForRequest(config *Config, req *http.Request, destination hostport
 		return decision
 	}
 
-	ACLDecision, err := config.EgressACL.Decide(role, destination.Host)
+	// X-Upstream-Https-Proxy is a header that can be set by the client to specify
+	// a _subsequent_ proxy to use for the CONNECT request. This is used to allow traffic
+	// flow as in: client -(TLS)-> smokescreen -(TLS)-> external proxy -(TLS)-> destination.
+	// Without this header, there's no way for the client to specify a subsequent proxy.
+	var connectProxyHost string
+	if len(req.Header["X-Upstream-Https-Proxy"]) > 0 {
```
cds2-stripe, line 943: Instead of indexing like this directly into the header map could we use https://pkg.go.dev/net/http#Header.Get

pkg/smokescreen/smokescreen.go
```diff
@@ -932,7 +935,18 @@ func checkACLsForRequest(config *Config, req *http.Request, destination hostport
...
+		connectProxyHost = req.Header["X-Upstream-Https-Proxy"][0]
```
cds2-stripe, line 944: We can reuse the value obtained above

pkg/smokescreen/smokescreen.go
```diff
@@ -932,7 +935,18 @@ func checkACLsForRequest(config *Config, req *http.Request, destination hostport
...
+	} else {
+		connectProxyHost = ""
```
cds2-stripe, line 946: We can remove this else block - the default value of a string in Go when it is initialized is an empty string.

pkg/smokescreen/testdata/acl.yaml
```diff
@@ -15,6 +15,20 @@ services:
   - name: test-open-srv
     project: security
     action: open
+  - name: test-external-connect-proxy-blocked-srv
+    project: security
+    action: enforce
+    allowed_domains:
+      - 127.0.0.1
+    allowed_external_proxies:
+      - myproxy.com
```
cds2-stripe, line 24: Could we add multiple external proxies to make sure the glob gets constructed correctly?

pkg/smokescreen/smokescreen_test.go
```diff
@@ -1236,6 +1236,77 @@ func TestCustomRequestHandler(t *testing.T) {
 	})
 }
 
+func TestCONNECTProxyACLs(t *testing.T) {
```
cds2-stripe, line 1239: Thanks for adding this!

pkg/smokescreen/smokescreen.go
```diff
@@ -642,8 +642,11 @@ func handleConnect(config *Config, pctx *goproxy.ProxyCtx) (string, error) {
 		return "", pctx.Error
 	}
 
-	// checkIfRequestShouldBeProxied can return an error if either the resolved address is disallowed,
-	// or if there is a DNS resolution failure.
+	/*
```
cds2-stripe, line 645: Just for consistency and since this is only 3 lines could we keep using the single line comment format?

cds2-stripe (dismissed):

pkg/smokescreen/smokescreen.go
pspieker-stripe: Fixed!

pkg/smokescreen/smokescreen.go
pspieker-stripe: Fixed as well

pkg/smokescreen/smokescreen.go
pspieker-stripe: Sure, fixed

pkg/smokescreen/testdata/acl.yaml
pspieker-stripe: This role is just used for the non-allowed proxies list; do you mean adding `localhost` here? How are you thinking of testing glob construction? Is just adding some different possible URL shapes helpful here, or is there something in particular you'd like us to assert against?

pspieker-stripe:
Not sure why `Test / test (1.21) (pull_request)` is failing now btw - seems flakey but not sure it's unrelated?

pkg/smokescreen/testdata/acl.yaml
cds2-stripe: All of the addresses here construct a glob which should match any `allowed_external_proxies` specific in the header. My thought was to add something like `myproxy2.com` to the end of the this and then verify it is still allowed as an external proxy in one of your tests.

cds2-stripe (approved):

PR #229, opened 2024-09-17
Use goveralls parallel build

harold-stripe:
# Description
Second attempt (https://github.com/stripe/smokescreen/pull/228) to try and fix the build
Following the [parallel builds](https://docs.coveralls.io/parallel-builds) pattern by setting the `COVERALLS_PARALLEL` env  and running `goveralls -parallel-finish=true`

# Test
Looks like it works in https://github.com/stripe/smokescreen/actions/runs/10911519945


cds2-stripe (approved):
Thank you!

Issue #236, opened 2024-10-18
--allow-ranges/--deny-address combination does not work as expected


eastebry:
Hello! Playing around with smokescreen, I noticed a potential rough edge. I'm using this command line configuration:

```
./smokescreen --listen-port 8888 --allow-range '192.168.0.0/16' --deny-address '192.168.1.5'
```
I would expect this configuration to deny access to `192.168.1.5`, but allow all other addresses in the 192.168 range. However, this is not what happens. The explicitly denied address is still allowed. This curl command: 

```
curl --proxy http://localhost:8888 http://192.168.1.5:8000/
```

results in these log lines:
```
{"allow":true,"conn_establish_time_ms":6,"content_length":928,"decision_reason":"Egress ACL is not configured","dns_lookup_time_ms":0,"enforce_would_deny":false,"id":"cs9eo7rjuspjjhc5lb70","inbound_remote_addr":"[::1]:57632","level":"info","msg":"CANONICAL-PROXY-DECISION","outbound_local_addr":"192.168.1.5:57633","outbound_remote_addr":"192.168.1.5:8000","proxy_type":"http","requested_host":"192.168.1.5:8000","start_time":"2024-10-18T23:15:43.082352Z","time":"2024-10-18T16:15:43-07:00","trace_id":""}
```

Interestingly, the inverse configuration using `--allow-address` and `--deny-range` works as expected. The below command allows requests to 93.184.215.14
```
./smokescreen --listen-port 8888 --deny-range 93.0.0.0/8 --allow-address 93.184.215.14
```

I'm hoping to have more time to dig into this next week. Perhaps the issue is that the allow list is always being given precedence over the deny list? If so, its probably important to document how these rules are applied. It makes more sense to have the more specific options (`--allow/deny-address`) to take precedence.

PR #237, opened 2024-10-18
Adding [allow|deny]_addresses settings to yaml config file

eastebry:
Pretty straightforward - adding the allow/deny command line arguments to the config loader so they can be included in a yaml configuration. This allows you to have a yaml config that looks like this:

```

ip: "127.0.0.1"
port: 8888

deny_ranges:
  - 93.0.0.0/8 

allow_addresses:
  - 93.184.215.14

allow_ranges:
  - 192.168.0.0/16

deny_addresses:
  - 192.168.1.5
```

I also opened a separate issue (https://github.com/stripe/smokescreen/issues/236) when playing around with this configuration, and discovering that the allow/deny configurations don't work as I would expect. This doesn't make a difference for this PR though.

### DIFF
--- a/pkg/smokescreen/config_loader.go
+++ b/pkg/smokescreen/config_loader.go
@@ -28,6 +28,8 @@ type yamlConfig struct {
 	Port                 *uint16
 	DenyRanges           []string `yaml:"deny_ranges"`
 	AllowRanges          []string `yaml:"allow_ranges"`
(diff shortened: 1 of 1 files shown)


jjiang-stripe (approved):
This seems pretty reasonable to me! It's nice to be able to apply these settings via a YAML config instead of only CLI

PR #239, opened 2024-12-10
Upgrade to `github.com/DataDog/datadog-go/v5`

LINKIWI:
This PR proposes upgrading the module's `datadog-go` dependency from v4 to the current major version, v5, with import path `github.com/DataDog/datadog-go/v5`.

There is only one relevant breaking change related to the namespace option which I have corrected inline. I also updated the vendored copy of the module with `go mod vendor`.


gauthamw-stripe:
Thanks for this! We missed on triaging this initially but have done this upgrade now - https://github.com/stripe/smokescreen/pull/319

PR #240, opened 2024-12-13
Fix CRL for client certificates

ne-bknn:
Hi! 

CRL feature does not seem to work. There are two main issues with it:
1) CRLs are loaded into `Config` structure, but **are not used anywhere**. I've implemented `VerifyPeerCertificate` function in config.TlsConfig
2) Even with implemented `VerifyPeerCertificate`, CRLs do not work since CRL loading occurs earlier, than CA loading. Smokescreen cannot find CA for CRL and fails. That's why I rearranged them.

I wrote an integration test with its own CA generation procedure - I did not want to figure how to use your test PKI in testdata, so I generate PKI on the fly and clean it up afterwards. 

This may look terrible, as it is just a draft. I'll clean it up, fix deprecation issues and resolve any of your comments if you are willing to merge it.

### DIFF
--- a/cmd/integration_test.go
+++ b/cmd/integration_test.go
@@ -18,6 +18,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"os/exec"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


ne-bknn:
Fixed in #293.

PR #253, opened 2025-07-07
Include status code in timeout response logs

rishikej-stripe:
### Summary
In this PR, I have included the response status_code inside the log entry object for HTTP error responses.

### Motivation
To analyze and log errors in http responses, we need various information, one of them being HTTP status code. Hence including its value in the log entry would help check for certain conditions, especially during logging.

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -366,6 +366,7 @@ func rejectResponse(pctx *goproxy.ProxyCtx, err error) *http.Response {
 			status = "Gateway timeout"
 			code = http.StatusGatewayTimeout
 			msg = "Timed out connecting to remote host: " + e.Error()
+			sctx.Logger = sctx.Logger.WithField("status_code", code)
 
 		} else if e, ok := err.(*net.DNSError); ok {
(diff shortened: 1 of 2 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -366,6 +366,7 @@ func rejectResponse(pctx *goproxy.ProxyCtx, err error) *http.Response {
 			status = "Gateway timeout"
 			code = http.StatusGatewayTimeout
 			msg = "Timed out connecting to remote host: " + e.Error()
+			sctx.Logger = sctx.Logger.WithField("status_code", code)
```
saurabhbhatia-stripe, line 369: Understanding the motivation here, currently hubble does not contain log like other than CANONICAL-PROXY-DECISION and CANONICAL-PROXY-CN-CLOSE, with this change would export additional log line for timeout requests .

Also, why are we selectively adding this to only gateway timeout, any reasons for not adding it any other error types.

cc: @manishsb-stripe incase you have context.

saurabhbhatia-stripe (approved):
LGTM.

PR #255, opened 2025-07-23
Populate log fields as soon as their values are known

rishikej-stripe:
### Notify
cc @stripe-internal/network-infra 
r? @manishsb-stripe 

### Summary

I have populated some log fields as soon as their context values are available, inside those functions which execute earlier in stage.

### Motivation

[Jira ticket.](https://jira.corp.stripe.com/browse/NET_INFRA-954)
In the logs, we see some fields are missing, which makes it difficult to understand and debug errors.

On investigation, I found that log fields are populated in 4 functions in smokescreen.go.
1. newContext()
2. dialContextLoggerFields()
3. logProxy()
4. extractContextLogFields()

The main reason behind this issue was that although the context was being added to the logs, in some edge cases the error logs were fired even before the log fields were populated. So I tried to populate the log fields as soon as their context values were available.

In this case, the fields "inbound_remote_x509_cn" and "inbound_remote_x509_ou" could be determined in the newContext() creation stage, hence I moved the fields from extractContextLogFields(). However the other condition inside extractContextLogFields() checking sctx.Decision != nil cannot be determined any earlier than calling checkIfRequestShouldBeProxied(), hence it cannot be moved.
Similarly I moved some fields from logProxy() to extractContextLogFields(), because they were dependent on decision and lookupTime variables, which were determined just before calling extractContextLogFields().

### Test plan

I added a test line err=errors.New("test") to simulate error just before line 669 inside handleConnect().
Before the changes:

Screenshot 2025-07-23 at 5 51 21 PM

After the changes:

Screenshot 2025-07-23 at 5 52 19 PM

- [ ] All changes in PR are covered by tests
- [x] Failures and edge cases tested

### Rollout/revert plan

  
  

  
  

Safe to revert unless specified otherwise

### Monitoring plan

- [ ] Change has no runtime impact
- [ ] Change is covered by automated alerts
- [ ] Change has other coverage (e.g. Splunk, Dashboards, Sentry, Queries, etc)

### DIFF
--- a/pkg/smokescreen/smokescreen.go
+++ b/pkg/smokescreen/smokescreen.go
@@ -617,9 +617,18 @@ func logProxy(pctx *goproxy.ProxyCtx) {
 
 	fields := logrus.Fields{}
 	decision := sctx.Decision
+	// If a lookup takes less than 1ms it will be rounded down to zero. This can separated from
+	// actual failures where the default zero value will also have the error field set.
+	fields[LogFieldDNSLookupTime] = sctx.lookupTime.Milliseconds()
 
 	if pctx.Resp != nil {
 		fields[LogFieldContentLength] = pctx.Resp.ContentLength
+	}
+
+	if sctx.Decision != nil {
+		fields[LogFieldDecisionReason] = decision.Reason
+		fields[LogFieldEnforceWouldDeny] = decision.enforceWouldDeny
+		fields[LogFieldAllow] = decision.allow
 	}
 
 	err := pctx.Error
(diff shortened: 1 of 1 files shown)


pkg/smokescreen/smokescreen.go
```diff
@@ -673,14 +675,14 @@ func handleConnect(config *Config, pctx *goproxy.ProxyCtx) (*goproxy.ConnectActi
 	// or if there is a DNS resolution failure, or if the subsequent proxy host (specified by the
 	// X-Https-Upstream-Proxy header in the CONNECT request to _this_ proxy) is disallowed.
 	sctx.Decision, sctx.lookupTime, pctx.Error = checkIfRequestShouldBeProxied(config, pctx.Req, destination)
+
+	// add context fields to all future log messages sent using this smokescreen context's Logger
+	sctx.Logger = sctx.Logger.WithFields(extractContextLogFields(pctx, sctx))
```
manishsb-stripe, line 680: This change is required for `handleHttp` as well. Can we have that too?

rishikej-stripe: For http case, this line is already there just after checkIfRequestShouldBeProxied() call, so I think no change is needed.

shubh-stripe (approved):

PR #256, opened 2025-07-24
Fix a race condition for prometheus metrics

sohamsen-stripe:
# Summary

A race condition occurs when multiple goroutines attempt to register the same Prometheus metric simultaneously, which causes a panic due to duplicate registration attempts.

Implemented a read-write mutex lock to avoid registering new metrics when they already exist.


saurabhbhatia-stripe (approved):
LGTM.

PR #257, opened 2025-07-25
Do not use example.com for testing

sohamsen-stripe:
# Summary

`example.com` has a few flaky IPs (the `96.7.*.*` IPs are not connecting), hence we removed its dependency from the tests.

For requests where we expect them to be blocked, we replace `127.0.0.1` with `stripe.com` as the latter is already present in `global_deny_list` in `testdata/acl.yaml`.

We have replaced the usage with a local test HTTP server.


saurabhbhatia-stripe (approved):
LGTM

PR #284, opened 2026-02-12
Implement tunnel limiting in smokescreen

rishikej-stripe:
### Summary
This PR does the following:

1. Adds a CONNECT tunnel limiting mechanism to prevent smokescreen resource exhaustion.
2. The tunnel limiter increments the acquired slot on every dialContext call and releases the slot when the connection is closed. This tunnel slot release is invoked upon InstrumentedConn closure, which happens when the tunnel is closed.

NOTE: This allows as many CONNECT requests to call smokescreen, perform ACL checks etc, but only the tunnel limit number of connections can actually call `DialContext()` and establish tunnel. If we want to limit pre-Dial access also, we can configure the rate-limiter and concurrency-limiter for that purpose.

### Motivation
The `max_concurrent_requests` configuration in Smokescreen is ineffective for CONNECT tunnels. The rate limiter acquires a slot when ServeHTTP() starts and releases it via defer when the function returns. For CONNECT requests, goproxy hijacks the connection and ServeHTTP() returns immediately while the tunnel continues in background goroutines. This allows unlimited CONNECT tunnels regardless of the configured limit, enabling resource exhaustion (OOM crash in ~10 seconds with 64MB memory limit).

The current `max_concurrent_requests` is basically limiting the number of connections being served by smokescreen, but does not monitor if they still last. We need a dedicated tunnel limiter for that.

### Test plan

We create 20 non-CONNECT requests and then 20 CONNECT requests to a slow backend via smokescreen proxy. The first 20 requests get handled by the existing concurrency limiter in smokescreen. The tunnel limiting restricts the number of CONNECT requests.

**Before**

We see the number of connections exceeding the `max_concurrent_requests` limit and blocking smokescreen resources for 5 seconds.

https://github.com/user-attachments/assets/1c36a20c-0ee9-4e38-919c-b513d88d0d8a

**After**
We see the tunnel limit is applied and allows only 3 requests. 17 other CONNECT requests are rejected

https://github.com/user-attachments/assets/a48a5225-d34e-407d-b0ec-fde1cbfed1f7

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,9 @@ Here are the options you can give Smokescreen:
    --max-request-rate value                    Maximum requests per second. 0 = unlimited (default: 0)
    --max-request-burst value                   Maximum burst capacity. Must be > max-request-rate when specified.
                                                  Omit to use default (2x max-request-rate).
+   --max-concurrent-connect-tunnels value      Maximum number of concurrent CONNECT tunnels.
+                                                 Unlike max-concurrent-requests, this limits actual long-lived connections.
+                                                 0 = unlimited (default: 0)
    --dns-timeout DURATION                      Maximum time to wait for DNS resolution (default: 5s)
    --version, -v                               print the version
 ```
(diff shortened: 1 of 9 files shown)


shubh-stripe (commented):
We shouldn't be wrapping in denyError now? The ideal behaviour still should return a 503 or 429.

shubh-stripe (approved):

PR #286, opened 2026-04-07
acl: pass *http.Request through Decide via DecideArgs

harold-stripe:
## Summary                                                                                                              

Aim was to add ``Req *http.Request`` as an arg to `ACL.Decide`:

I replaced the `Decider` interface's three positional string arguments with a `DecideArgs` struct:                         
                                                                                                                          
  ```go                                                                                                                   
  // Before                                                                                                               
  Decide(service, host, connectProxyHost string) (Decision, error)                                                        
                                                                                                                          
  // After
  Decide(args DecideArgs) (Decision, error)                                                                               
```
  DecideArgs carries `Service`, `Host`, `ConnectProxyHost` and a new `Req *http.Request` field. `ACL.Decide` does not inspect Req, it is available solely for custom implementations.
                                                                                                                          
I also removed `TestHostSquareBrackets` (unreachable since Go 1.21 rejects bracket hostnames at both the HTTP client and    
  server layers before requests reach smokescreen's handler).                             
                                                                                                                        
##  Motivation

Custom Decider implementations need access to the full HTTP request: `headers`, context, caller identity to make policy decisions beyond what service name and hostname provide. Adding `*http.Request` as a fourth positional argument would make the signature unwieldy and every future extension a breaking change. A struct keeps the interface stable.          

##  Test plan

- `TestDecideReqIgnored`: new test asserting `ACL.Decide` produces identical results regardless of the `Req` value            
- `TestDecisionWithProxyHost`: extended with a missing case: allowed proxy + disallowed target (should deny)
- `TestInvalidUpstreamProxyConfiguratedFromEnv`: now uses nonexistent.invalid as target, deterministic across environments
- CI passes

### DIFF
--- a/pkg/smokescreen/acl/v1/acl.go
+++ b/pkg/smokescreen/acl/v1/acl.go
@@ -2,14 +2,26 @@ package acl
 
 import (
 	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 	"github.com/stripe/smokescreen/pkg/smokescreen/hostport"
 )
 
+// DecideArgs holds the arguments for an ACL decision. Using a struct keeps the
+// Decider interface stable as new fields are added (e.g. Req for request-aware
+// authorization in custom Decider implementations). ACL.Decide does not
+// inspect Req; it is provided solely for use by custom implementations.
+type DecideArgs struct {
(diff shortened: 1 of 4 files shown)


rishikej-stripe (approved):
LGTM, thanks!

PR #293, opened 2026-06-18
Fix CRL enforcement for client certificate revocation

abhishoya-stripe:
### Background

The `--tls-crl-file` flag loads and parses a Certificate Revocation List but never actually enforces it during TLS handshakes. This is due to two bugs:

1. CRL never checked at runtime - The parsed CRL is stored in but no VerifyConnection callback existed to check client certificates against it. A client presenting a revoked certificate was accepted identically to one with a valid certificate
2. CLI initialization ordering - SetupCrls() uses a client CA Map which is setup as part of SetupTls(). Since SetupCrls() ran first, it always found the map empty, and ignored the CRLs

### Summary

- pkg/smokescreen/config.go: Added a VerifyConnection callback to the TLS config that checks each client certificate's serial number against the loaded CRL
- cmd/smokescreen.go: Moved SetupCrls() to after SetupTls() so the CA map is populated when CRL signatures are verified
- cmd/integration_test.go: Added TestCRLEnforcement verifying that valid certs are accepted and revoked certs are rejected at the TLS layer
- cmd/testdata/pki/: Regenerated test PKI with proper Authority Key Identifier extension on the CRL (required for CRL-to-CA matching)

### Impact

- Deployments using --tls-crl-file will now actually enforce revocation. Previously, this flag was non-functional — revoked certificates were silently accepted
- Deployments not using --tls-crl-file are unaffected (the callback short-circuits when no CRLs are loaded)


pkg/smokescreen/config.go
```diff
@@ -626,6 +626,32 @@ func (config *Config) SetupTls(certFile, keyFile string, clientCAFiles []string)
...
+					issuerKeyId := string(cert.AuthorityKeyId)
+					crl, ok := config.CrlByAuthorityKeyId[issuerKeyId]
+					if !ok {
+						continue
+					}
+
+					for _, revoked := range crl.TBSCertList.RevokedCertificates {
+						if cert.SerialNumber.Cmp(revoked.SerialNumber) == 0 {
+							return fmt.Errorf("certificate with serial %s has been revoked", cert.SerialNumber.String())
+						}
```
shubh-stripe: Should we do a linear scan each time? I guess we can take some memory hit and create a map out of this for faster lookups? 
This would add up fast for each request if the CRL is large.

pkg/smokescreen/config.go
```diff
@@ -626,6 +626,32 @@ func (config *Config) SetupTls(certFile, keyFile string, clientCAFiles []string)
...
 		ClientAuth:   clientAuth,
 		ClientCAs:    clientCAs,
+		VerifyConnection: func(cs tls.ConnectionState) error {
+			if len(config.CrlByAuthorityKeyId) == 0 {
+				return nil
+			}
+
+			for _, chain := range cs.VerifiedChains {
+				for _, cert := range chain {
+					if cert.IsCA {
```
shubh-stripe, line 645: this would also skip intermediate certs? is that expected form a smokescreen POV (clients can have full chain of certs)

shubh-stripe (commented):
minor doubts and improvements

pkg/smokescreen/config.go
abhishoya-stripe: Yeah, it will skip intermediate certs. That's expected because we only check leaf client certificates against the CRL. Intermediate CA revocation should be handled by removing the CA from the ClientCAs trust pool rather than via CRL. 

Intermediates in VerifiedChains are there because Go includes the full chain up to the root, but they aren't the certs we're revoking.

pkg/smokescreen/config.go
abhishoya-stripe: I have updated the logic to build and use a map now

shubh-stripe (approved):

PR #297, opened 2026-07-22
Fix connect timeout defaults

gauthamw-stripe:
## Summary
Ensure we set the default timeout as expected.

Issue 1: The Config obj is initialized with timeout 0. This is a miss for users not loading a config file.
Issue 2: The config yaml loader expects a non-nil ConnectTimeout so if the yaml doesn't specify a value, the loader sets it to 0 and that overrides the Config object, making it unlimited.

## Validation
Wrote custom unit-test to validate this.

### DIFF
--- a/pkg/smokescreen/config_loader.go
+++ b/pkg/smokescreen/config_loader.go
@@ -38,7 +38,7 @@ type yamlConfig struct {
 	AllowMissingRole     bool     `yaml:"allow_missing_role"`
 	Network              string   `yaml:"network"`
 
-	ConnectTimeout time.Duration  `yaml:"connect_timeout"`
+	ConnectTimeout *time.Duration `yaml:"connect_timeout"`
 	IdleTimeout    time.Duration  `yaml:"idle_timeout"`
 	ExitTimeout    *time.Duration `yaml:"exit_timeout"`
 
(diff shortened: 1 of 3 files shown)


kshitizarora-stripe (approved):

pkg/smokescreen/config_loader.go
```diff
@@ -38,7 +38,7 @@ type yamlConfig struct {
 	AllowMissingRole     bool     `yaml:"allow_missing_role"`
 	Network              string   `yaml:"network"`
 
-	ConnectTimeout time.Duration  `yaml:"connect_timeout"`
+	ConnectTimeout *time.Duration `yaml:"connect_timeout"`
```
nishants-stripe, line 41: Technically, this is a backward incompatible change. Users who had connect_timeout unset, had unbounded connection time. After this change, they will have a 10s connect timeout. I understand this is intentional, but may be worth calling out

gauthamw-stripe: We've not released smokescreen in a while so users today are pinned to an old release unless they manually pin a newer SHA (ie `latest` won't pick this). I plan to publish a backward-incompatibility note/guide with a new release once all my planned changes are in.

kshitizarora-stripe (approved):

PR #299, opened 2026-08-19
Pin GitHub Actions to commit SHAs

sohamsen-stripe:
Pins third-party GitHub Actions to immutable commit SHAs instead of mutable version tags. This helps prevent supply chain attacks by ensuring workflow dependencies cannot change without a reviewed update.


anuragm-stripe (approved):
LGTM

